PR #4, opened 2018-02-04
fix issue-3

ljgw:
Changes in this branch ensure
1. Multiple *produces* list-items result in the `hasMore` in all of them, except the last. (so, comma's in the correct locations in the generated code)
2. Only distinct media-types in *produces*

### DIFF
--- a/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
@@ -1835,6 +1835,11 @@ public abstract class DefaultCodegenConfig implements CodegenConfig {
                     codegenOperation.getVendorExtensions().put(CodegenConstants.IS_RESPONSE_FILE_EXT_NAME, Boolean.TRUE);
                 }
             }
+            if (codegenOperation.produces != null){
+                Set<String> mediaTypes = new HashSet<String>();
+                codegenOperation.produces.removeIf(map -> !mediaTypes.add(map.get("mediaType")));
+                codegenOperation.produces.get(codegenOperation.produces.size() - 1).remove("hasMore");
+            }
             codegenOperation.responses.get(codegenOperation.responses.size() - 1).getVendorExtensions().put(CodegenConstants.HAS_MORE_EXT_NAME, Boolean.FALSE);
 
             if (methodResponse != null) {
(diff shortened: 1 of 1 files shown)


ljgw:
travis fails now - as far as I can see it's because the necessary version of guava (23) was updated in the swagger-codegen/pom.xml on the 17-01-2018, but travis seems to be using a snapshot of a few days earlier: 

> [INFO] Downloading from sonatype-snapshots: https://oss.sonatype.org/content/repositories/snapshots/io/swagger/swagger-codegen/3.0.0-SNAPSHOT/swagger-codegen-3.0.0-20180112.231844-20.pom

jmini:
@ljgw the problem is that `guava` no longer in the dependency tree is.

I have compared the last released version (`1.0.0-rc0`) and the current SNAPSHOT version.

Comparison of the dependency tree

`guava` version `23.0` was provided by `swagger-codegen` (version `3.0.0-rc0`). With the current `3.0.0-SNAPSHOT` version, this is no longer the case.

See issue #5

jmini:
You can rebase your change on top of PR #6, but as I wrote there this only solves the first errors mentionned in your Travis Build:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.6.1:compile (default-compile) on project swagger-codegen-generators: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/swagger-api/swagger-codegen-generators/src/main/java/io/swagger/codegen/languages/java/AbstractJavaCodegen.java:[4,30] package com.google.common.base does not exist
[ERROR] /home/travis/build/swagger-api/swagger-codegen-generators/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java:[4,30] package com.google.common.base does not exist
[ERROR] /home/travis/build/swagger-api/swagger-codegen-generators/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java:[5,33] package com.google.common.collect does not exist
[ERROR] /home/travis/build/swagger-api/swagger-codegen-generators/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java:[63,24] cannot find symbol
```

jmini:
now that #7 is fixed, rebasing on master should help your change

ljgw:
Thanks, @jmini. travis-ci will still fail because it cannot find  `./mvnw`, but at least everything compiles now on travis-ci.

jmini:
Some code was pushed on master for `mvnw`(see #8)

ljgw:
Some more code was pushed on master to fix the travis-ci build, and now it passes! I rebased on master and pushed.

HugoMario:
thanks @ljgw and @jmini i just merged the PR.

Issue #5, opened 2018-02-05
Migrate guava logic to java 8


HugoMario:
There is an error with dependency classes from guava. But instead fixing it with maven dependency declaration, let's take this chance to migrate that logic to Java 8.

jmini:
At the moment there is a compile dependency on Guava:

* In [AbstractJavaCodegen#sanitizePackageName(String)](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/languages/java/AbstractJavaCodegen.java#L1098) (line 1098) you are using guava perform this check: `Strings.isNullOrEmpty(packageName)`
* In [DefaultCodegenConfig.removeNonNameElementToCamelCase(String, String)](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java#L2883) (line 2883) both `com.google.common.base.Function` and `com.google.common.collect.Lists` are used.

@HugoMario: I guess you mention rewritting the code for the second use case.

---

In the mean time I had a look a the dependencies tree, guava `15.0` is already a `test` scope dependency, so I added the same version a `compile` dependencies.

```
<dependency>
	<groupId>com.google.guava</groupId>
	<artifactId>guava</artifactId>
	<version>15.0</version>
</dependency>
```

jmini:
I think the guava version included in the last released version of this project (`1.0.0-rc0`) was guava `23.0`.

The guava dependency is no longer present in the last SNAPSHOT version used in the current dependencies.

See also my analysis (with screenshot) in #4.

jmini:
I have missed usage of `com.google.common.collect.LinkedListMultimap` in `JavaClientCodegen`. Not sure what you want to do about it.

HugoMario:
Well, the idea is remove any `guava` dependency

jmini:
To replace `LinkedListMultimap` with java8 code, I found this blog post: http://fahdshariff.blogspot.fr/2017/07/java-8-multimaps.html

Second pull request: #7.

HugoMario:
Travis has a success build now, but it seems there is a config error.
https://api.travis-ci.org/v3/job/338265006/log.txt

PR #6, opened 2018-02-06
remove dependency to guava in code

jmini:
This is a possible fix for #5.

With this change, the java code no longer uses guava. Imports of following classes are removed:

* `com.google.common.base.Function`
* `com.google.common.collect.Lists`
* `com.google.common.base.Strings`

This change do not compile on my machine, because other changes occured in the SNAPSHOT libraries referenced as dependency.

In oder to verify my change, I had to switch back to the same released versions as with release `1.0.0-rc0`, properties:

```
<swagger-parser-version>2.0.0-rc3</swagger-parser-version>
<swagger-core-version>2.0.0-rc4</swagger-core-version>
<swagger-codegen-version>3.0.0-rc0</swagger-codegen-version>
```

Changing the versions is not a solution that should be commited to the master branch. The code needs to be changed to work with the lattest SNAPSHOT versions

### DIFF
--- a/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
@@ -1,8 +1,6 @@
 package io.swagger.codegen.languages;
 
 import com.github.jknack.handlebars.Handlebars;
-import com.google.common.base.Function;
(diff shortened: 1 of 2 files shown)


HugoMario:
thanks @jmini it does still not fix #5 , but i'll work on it.

PR #7, opened 2018-02-06
remove guava dependency (2)

### LINKED ISSUE #5: Migrate guava logic to java 8

jmini:
One more step to fix #5.
Second step for #6.


HugoMario:
Thanks @jmini

Issue #20, opened 2018-02-22
collection-fomat query param template rendering failed for retrofit 1 & 2


romaindequidt:
Let's try to generate Java client for this OpenAPI definition file:
```
openapi: 3.0.0
paths:
  /test:
    get:
      tags:
        - Test
      summary: Test
      parameters:
        - $ref: '#/components/parameters/IDList'
      responses:
        '200':
          description: Test
components:
  parameters:
    IDList:
      name: ids
      description: |
        List of ids.
      in: query
      style: form
      schema:
        type: array
        items:
          type: string
```
The current templates:
`{{#is this 'query-param'}}@retrofit.http.Query("{{baseName}}") {{#if collectionFormat}}{{#is this 'collection-format-multi'}}{{{dataType}}}{{/is}}{{#isNot this 'collection-format-multi'}}{{{collectionFormat.toUpperCase}}}Params{{/isNot}}{{else}}{{{dataType}}}{{/if}} {{paramName}}{{/is}}`
`{{#is this 'query-param'}}@retrofit2.http.Query("{{baseName}}") {{#if collectionFormat}}{{#is this 'collection-format-multi'}}{{{dataType}}}{{/is}}{{#isNot this 'collection-format-multi'}}{{{collectionFormat.toUpperCase}}}Params{{/isNot}}{{else}}{{{dataType}}}{{/if}} {{paramName}}{{/is}}`
generates the foloowing output:
```
@retrofit2.http.Query("ids") Params ids
```
Wich produces the following errot at complie time:
```
Error:(42, 34) java: cannot find symbol
  symbol:   class Params
  location: interface io.swagger.client.api.TestApi
```
CSVParams or List<String> should be a better dataType

My pom.xml is the folowing:
```
<plugin>
    <groupId>io.swagger</groupId>
    <artifactId>swagger-codegen-maven-plugin</artifactId>
    <version>${swagger-codegen.version}</version>
    <executions>
        <execution>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputSpec>src/main/resources/openapi/openapi.yaml</inputSpec>
                <language>java</language>
                <library>retrofit2</library>
                <configOptions>
                    <dateLibrary>java8</dateLibrary>
                </configOptions>
                <generateApiTests>false</generateApiTests>
            </configuration>
        </execution>
    </executions>
    <dependencies>
        <dependency>
            <groupId>io.swagger</groupId>
[...]

PR #21, opened 2018-02-22
Fix collection-fomat query param datetype templates

romaindequidt:
As explained in https://github.com/swagger-api/swagger-codegen-generators/issues/20, collection-fomat query param template rendering failed for retrofit 1 & 2.

I don't knw if it's the good solution.
I just make queryParam templates work (because it's not the case so far) but I don't know if it's the good/best solution.

In another issue (https://github.com/swagger-api/swagger-codegen/issues/6504), they'd prefer to improve "pathParam" template to manage much more cases.


HugoMario:
thanks a lot @romaindequidt!

PR #24, opened 2018-02-25
Map description from RequestBody in CodegenParameter

jmini:


HugoMario:
looks good, thanks @jmini !

PR #32, opened 2018-03-07
Remove jmustache dependency

jmini:


jmini:
After the change made in PR https://github.com/swagger-api/swagger-codegen/pull/7757, this project no longer depends from `jmustache`.

PR #41, opened 2018-03-15
Migrate `html2` Codegen

jmini:
Migration of the `html2` generators

According to the step by step procedure described here: [Swagger Codegen migration](https://github.com/swagger-api/swagger-codegen/wiki/Swagger-Codegen-migration-(swagger-codegen-generators-repository))

Clean-up tasks for the `swagger-codegen` repository are prepared on my branch [patch-12](https://github.com/jmini/swagger-codegen/tree/patch-12).
I will open the pull request once this pull request is merged.

Known issue:

* Blank spaces in the generated html code (handlebars is used instead of mustache, fine-tuning of the templates can be made later)

PR #42, opened 2018-03-16
Fix compile error

jmini:
Introduced due to a conflict between #40 and #41.

@HugoMario can you merge this?


jmini:
@HugoMario has pushed a similar fix with db4b4920c00d779652d59e0580685da04ca2c73a

PR #43, opened 2018-03-16
Allow some values to be use in templates

jmini:
Existing templates (here or in the `swagger-codegen` repository) are using these values:

```
{{modelPackage}}
{{apiPackage}}
{{invokerPackage}}
{{#sortParamsByRequiredFlag}}..{{/sortParamsByRequiredFlag}}
{{#hideGenerationTimestamp}}..{{/hideGenerationTimestamp}}
```

They need to be available in the `additionalProperties` map in order to be consumed from the `mustache` templates.

In addition, setting the value should be possible using setters. This is usefull when the generators are used directly (no maven plugin or no cli integration). Setter for `hideGenerationTimestamp` has been added.

`invokerPackage` is present only in `AbstractJavaCodegen`. Its value is derived from `apiPackage` or `modelPackage`. Current behavior needs to be preserved this is why some regression tests were implemented first in 7ad6ca24373e71f421aa96a5e82031cf33fb7d22.

### DIFF
--- a/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
@@ -166,23 +166,23 @@ public abstract class DefaultCodegenConfig implements CodegenConfig {
 
         if (additionalProperties.containsKey(CodegenConstants.MODEL_PACKAGE)) {
             this.setModelPackage((String) additionalProperties.get(CodegenConstants.MODEL_PACKAGE));
-        } else {
-            // not set, use to be passed to template
+        } else if (StringUtils.isNotEmpty(modelPackage)) {
+            // not set in additionalProperties, add value from CodegenConfig in order to use it in templates
             additionalProperties.put(CodegenConstants.MODEL_PACKAGE, modelPackage);
         }
 
         if (additionalProperties.containsKey(CodegenConstants.API_PACKAGE)) {
             this.setApiPackage((String) additionalProperties.get(CodegenConstants.API_PACKAGE));
-        } else {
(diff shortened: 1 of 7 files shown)


src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
```diff
@@ -166,15 +166,24 @@ public void processOpts() {
 
         if (additionalProperties.containsKey(CodegenConstants.MODEL_PACKAGE)) {
             this.setModelPackage((String) additionalProperties.get(CodegenConstants.MODEL_PACKAGE));
+        } else {
+            // not set, use to be passed to template
+            additionalProperties.put(CodegenConstants.MODEL_PACKAGE, modelPackage);
```
HugoMario: I don't think this could cause an error or something similar, but i'm wondering if you'd need to set modelPackage in additionalProperties if its value is null or empty.

jmini:
Thank you for your feedback. I have added a _is null or empty_ check. 

To stay consistent, I updated all code location where the same pattern is used.

HugoMario:
awesome, thanks !

HugoMario:
@jmini 
i will need to revert these PR, since is causing problem to php client generator, and i think it'll be present to next languages generators to add.

The best approach here is add those changes in the specific language overriding methods in `DefaultCodegenConfig` class, instead changing it.

I understand is more work, i can help you with that if you're agree.

HugoMario:
basically i will only rever this [commit](https://github.com/swagger-api/swagger-codegen-generators/pull/43/commits/e238153d8134e213d83e29cf6d956b98f1750aab) and tweaks test related.

PR #65, opened 2018-04-19
Added Swift OpenAPI 3.0 generators.

koresak:
I've prepared an adaptation of Swift 3 and Swift 4 code generators for OpenAPI 3.0. Unfortunately there are some open point yet because I haven't got into the topic so deep...

- Copy of tests are not adapt/working on 100%. Check the test methods io.swagger.codegen.languages.swift.Swift3CodegenTest#binaryDataTest and io.swagger.codegen.languages.swift.Swift4CodegenTest#binaryDataTest. There are two TODOs with the comments. I guess that a behavior of the CodegenOperation might have changed, but not sure...
- I had to copied and adapt binaryDataTest.json and datePropertyTest.json files in order to have all the tests prepared. But not sure if it is wanted to have these JSON files in the src/test/resources directory... These files are converted to the OpenAPI 3.0 spec. But:
- I had to add the section `,
  "components": {
    "schemas": {
      "Some": {
        "properties": {
          "id": {
            "type": "integer",
            "format": "int64"
          },
          "name": {
            "type": "string"
          }
        }
      }
    }
  }` at the end of the JSON files. This might be misunderstanding of the original tests...
- I have no idea how to set the proper version in the generated .swagger-codegen/VERSION file. The current value is "unset".
- The generated code contains such a lot of empty lines in comparison with the swagger-codegen 2.x. I need a hint where to add ~...
- The generated code is missing examples. I don't know why.
- The generated Models contains Body.swift and Body1.swift files without any reason. This is common for all the generators.
- I don't know if {{#allVars.isEmpty}} in mustache files should be replaced. If they do, I need a hint how to do it.


HugoMario:
thanks a lot @koresak !!

About your topics:

- Those tests with TODO's is because they need to be tweaked to the new changes according the oas3 api in most of the cases.
- It's good to have oas 3 definition, but keep in mind that codegen v3 supports both format. Well the conversion is done by parser.
- About `components/schemas` section, that's something that actually needs to be place on tests instead `definitions`.
- i would need to check the issue you mention about `VERSION`
- The `~` chars at the end of elements `{{element~}}` specially on close elements.
- I think the example is a present issue that needs to be addressed.

I'll try to help with version / and example issues and let you know.

koresak:
@HugoMario I have a question regarding the empty lines topic:
What is our priority? Having similar output as the output from swagger-codegen 2.x or readable mustache files?

Because there is a difference between those samples:
`    /**{{#summary}}`
`     {{{summary}}}`
`    {{/summary~}}`
And the second one:
`    /**`
`    {{#summary}}`
`     {{{summary}}}`
`    {{/summary~}}`
The first one generates same output as the old codegen but the `    {{#summary}}` element has to be put on the same line as the beginning of the comment characters. The second example is the same code section as it was written in the old templates. How should I fix mustache templates?

koresak:
Well, @HugoMario I've just updated mustache templates in order to have output as similar as it is possible. The mustache files are less readable but the output is more or less almost same from the point of empty lines. There are still some differences but it is up to you if it is OK or if templates should be corrected a bit...

I would like to summarize known differences/open point in order to have an overview:

- [ ] **Swift 2 templates** Are you interested in this feature in the new oas 3 code generator? I skipped it because of the comment in the original class io.swagger.codegen.languages.SwiftCodegen:
`/**`
` * Swift (2.x) generator is no longer actively maintained. Please use `
` * 'swift3' or 'swift4' generator instead.`
` */`
- [ ] **Difference in generated code** - I am not sure if the generated code differences are caused by the difference in json files (v2.0 vs. v3.0) or some bug in the SwiftXCodegen class. Could you check the generated output?
- [ ] **.swagger-codegen/VERSION** file doesn't contain the correct version.
- [ ] generated code is missing **examples** in the comments.

PR #68, opened 2018-04-26
Use Content-Type and Accept headers only if defined

fabianhenniges:
The feign java client did send empty Content-Type and Accept
headers. JAX-RS spring boot server answered with a 400.


HugoMario:
thanks a lot @fabianhenniges !!

PR #72, opened 2018-04-30
Angular typescript client

kalexmills:
Migrated Angular typescript client. Model test works, but integration test seems to be hard to fix.

I would have taken care of the integration tests as well, but I am finding the filesystem path resolution mechanisms used in this repo rather confusing. In several places path separators are hard-coded to '/' characters. :-\

See #70 for further comments on referencing files on the resources path.


HugoMario:
thanks @kalexmills !

I have a doubt about field `isResponseTypeFile` on mustaches templates, i just don't see where is set/used on classes.

PR #73, opened 2018-05-01
Add close tag to HTML anchors

tomer:
According to HTML specs, <a> MUST HAVE a closing tag, so `<a name="foo" />` is invalid syntax. Fortunately, browsers switch to quirks mode when facing this issue, so there is appears to be working well, but the browsers won't behave as expected sometimes, for example if removing the link to top which follows this anchor two lines below. 

More information: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a


HugoMario:
thanks @tomer !

PR #84, opened 2018-05-28
Retrofit: Query Params generates a Path annotation

loicsiret:
Before this commit, the generator create a @retrofit2.http.Path for a query parameter. This path replace the @retrofit2.http.Query by a @retrofit2.http.Path


HugoMario:
thanks a lot @loicsiret sorry for this looooooong delay :/

PR #89, opened 2018-06-05
Migrating #has statement in api.mustache of JavaSpring from Mustache to Handlebars.

thomas-worm-datev:


HugoMario:
thanks a lot @thomas-worm-datev !!

Issue #91, opened 2018-06-08
[Spring] Duplicate "is" in boolean getter method names


Gabe-Linux:
When running codegen on schemas using the Spring language, boolean getter method names contain the "is" prefix twice.

For example, this model:
```yaml
Example:
      type: object
      properties:
        myprop:
          type: boolean
          default: True
```

Will generate this class:
```java
public class Example   {

  @JsonProperty("myprop")
  private Boolean myprop = true;

  public Example myprop(Boolean myprop) {
    this.myprop = myprop;
    return this;
  }

  /**
   * Get myprop
   * @return myprop
  **/
 
  @ApiModelProperty(value = "")
  public Boolean isisMyprop() {
    return myprop;
  }

  public void setMyprop(Boolean myprop) {
    this.myprop = myprop;
  }

  @Override
  public boolean equals(java.lang.Object o) {
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }
    Example example = (Example) o;
    return Objects.equals(this.myprop, example.myprop);
  }

  @Override
  public int hashCode() {
    return Objects.hash(myprop);
  }

  @Override
  public String toString() {
    StringBuilder sb = new StringBuilder();
    sb.append("class Example {\n");
    
    sb.append("    myprop: ").append(toIndentedString(myprop)).append("\n");
    sb.append("}");
    return sb.toString();
  }

  /**
   * Convert the given object to string with each line indented by 4 spaces
   * (except the first line).
   */
  private String toIndentedString(java.lang.Object o) {
    if (o == null) {
      return "null";
    }
    return o.toString().replace("\n", "\n    ");
  }
}
```

Notice the `isisMyprop()` method, which should be named `isMyprop()`.

nardil:
This issue is open for JavaJaxRs
```java
 /**
   * Tells if the resource is enabled
   * @return enabled
  **/
  @JsonProperty("enabled")
  public Boolean isisEnabled() {  // <<<< Should be isEnabled
    return enabled;
  }

  public void setEnabled(Boolean enabled) {
    this.enabled = enabled;
  }

  public Entity enabled(Boolean enabled) {
    this.enabled = enabled;
    return this;
  }
```

PR #92, opened 2018-06-08
Remove duplicate "is" as described in #91

Gabe-Linux:


HugoMario:
LGTM, thanks a lot @Gabe-Linux !!

PR #96, opened 2018-06-12
Several fixes

thomas-worm-datev:


HugoMario:
awesome, thanks @thomas-worm-datev !

PR #107, opened 2018-06-18
[Spring] Add @JsonTypeId annotation to fields used as discriminator property

### LINKED ISSUE #105: [Spring] @JsonTypeId not set on discriminator property causing serialization issues
Spring codegen doesn't add set a `@JsonTypeId` annotation on the field used as the discriminator property, causing duplications of this field when serializing to JSON.

NOTE: This example assumes issue #103 as fixed.

Using these schemas:

```yaml
Example:
      type: object
      required:
        - baseType
      discriminator:
        propertyName: baseType
      properties:
        baseType:
          type: string
        otherProp:
          type: string
SubExample:
      allOf:
        - $ref: '#/components/schemas/Example'
        - type: object
```

Will generate this `Example` class (truncated for illustration):

```java
/**
 * Example
 */
@Validated

@javax.annotation.Generated(value = "io.swagger.codegen.languages.java.SpringCodegen", date = "2018-06-18T16:55:52.608-04:00[America/New_York]")

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "baseType", visible = true )
@JsonSubTypes({
  @JsonSubTypes.Type(value = SubExample.class, name = "SubExample"),
})

public class Example   {

  @JsonProperty("baseType")
  private String baseType = null;
  
  @JsonProperty("otherProp")
  private String otherProp = null;
}
```

When trying to serialize an instance of this class as JSON, the discriminator field will be included twice, like so:

```json
{
    "baseType":"SubExample",
    "baseType":"SubExample",
    "otherProp":"some value"
}
```

## A solution
The `@JsonTypeId` annotation can be used to fix this. By declaring the annotation on the same field used as the discriminator, like so:

```java
public class Example   {

  @JsonProperty("baseType")
  @JsonTypeId
  private String baseType = null;
  
  @JsonProperty("otherProp")
  private String otherProp = null;
```

An instance of this class would properly serialize to JSON as:

```json
{
    "baseType":"SubExample",
    "otherProp":"some value"
}
```

Gabe-Linux:
Fixes #105

I attempted to follow the pattern already in use with using `vendorExtensions` and `#is` for additional features, while keeping as much logic as possible out of the handlebars template.

Please let me know if I took the incorrect approach here.

### DIFF
--- a/src/main/java/io/swagger/codegen/languages/java/AbstractJavaCodegen.java
+++ b/src/main/java/io/swagger/codegen/languages/java/AbstractJavaCodegen.java
@@ -372,6 +372,7 @@ public abstract class AbstractJavaCodegen extends DefaultCodegenConfig {
         importMapping.put("JsonProperty", "com.fasterxml.jackson.annotation.JsonProperty");
         importMapping.put("JsonSubTypes", "com.fasterxml.jackson.annotation.JsonSubTypes");
         importMapping.put("JsonTypeInfo", "com.fasterxml.jackson.annotation.JsonTypeInfo");
+        importMapping.put("JsonTypeId", "com.fasterxml.jackson.annotation.JsonTypeId");
         importMapping.put("JsonCreator", "com.fasterxml.jackson.annotation.JsonCreator");
         importMapping.put("JsonValue", "com.fasterxml.jackson.annotation.JsonValue");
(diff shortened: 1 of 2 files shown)


HugoMario:
LGTM, thanks @Gabe-Linux !!

PR #108, opened 2018-06-26
Issue 8330 - use consistent class name for parents so java class EXTENDS works properly

n24005:
Use toModelName to insure consistency in class names


HugoMario:
Hey @n24005 , thanks for your PR. it is possible to add tests to check this change? also we need to make sure it does not affect other "parent" names formats.

HugoMario:
adding this comment only to match with [#8330](https://github.com/swagger-api/swagger-codegen/issues/8330) on codegen repo:

HugoMario:
thanks @n24005, i checked this with various sample and it works ok, thanks a lot!

PR #115, opened 2018-07-07
added support for bearer security schema.

HugoMario:


src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
```diff
@@ -2518,7 +2518,11 @@ public boolean isDataTypeFile(String dataType) {
                 codegenSecurity.getVendorExtensions().put(CodegenConstants.IS_KEY_IN_QUERY_EXT_NAME, !isKeyInHeader);
 
             } else if (SecurityScheme.Type.HTTP.equals(schemeDefinition.getType())) {
-                codegenSecurity.getVendorExtensions().put(CodegenConstants.IS_BASIC_EXT_NAME, Boolean.TRUE);
+                if ("bearer".equalsIgnoreCase(schemeDefinition.getScheme())) {
+                    codegenSecurity.getVendorExtensions().put(CodegenConstants.IS_BEARER_EXT_NAME, Boolean.TRUE);
+                } else {
```
lorenzleutgeb: Here you assume that any authentication scheme that is not the bearer authentication scheme is the basic authentication scheme. This is wrong. [There are some others: Digest, HOBA, Mutual.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#Authentication_schemes). These will fail the first condition in line 2521, since they are not the bearer scheme, but they will also not match the intentions of the second case, namely that they are the basic scheme. You should treat the basic scheme as a special case and then as a first step, all the others in a generic way.

src/main/resources/v2/Java/ApiClient.mustache
```diff
@@ -103,6 +103,7 @@ public class ApiClient {
     authentications = new HashMap<String, Authentication>();{{#authMethods}}{{#is this 'basic'}}
     authentications.put("{{name}}", new HttpBasicAuth());{{/is}}{{#is this 'api-key'}}
     authentications.put("{{name}}", new ApiKeyAuth({{#is this 'key-in-header'}}"header"{{/is}}{{#isNot this 'key-in-header'}}"query"{{/isNot}}, "{{keyParamName}}"));{{/is}}{{#is this 'oauth'}}
+    authentications.put("{{name}}", new OAuth());{{/is}}{{#is this 'bearer'}}
```
lorenzleutgeb: Now this is fun, since there is also a security scheme type for OAuth in particular. I would expect that to generate an OAuth authentication. Getting an OAuth authentication when one wanted a plain bearer based authentication might be confusing.

HugoMario: Right, basically just picked the `OAuth` class because it add the `Authentication` header which we're looking for. I guess to avoid confuse we could create another class. But i'm open to ideas if you have something in mind.

Thanks again for your time to check this.

src/main/java/io/swagger/codegen/languages/DefaultCodegenConfig.java
HugoMario: thanks for your support @lorenzleutgeb. Mainly i added this code based that we're not supporting the other one, but will follow your idea.

Issue #122, opened 2018-07-14
[okhttp-gson] GSON config does not include discriminator type adapters


motorro:
Hi!

2.x [implementation](https://github.com/swagger-api/swagger-codegen/blob/e2c58fad717a9f802cc814b7dabc573fed78509c/modules/swagger-codegen/src/main/resources/Java/JSON.mustache#L63) sets up adapters based on discriminator. The [new one](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/resources/v2/Java/libraries/okhttp-gson/JSON.mustache#L52) does not.
Is it made on purpose?

PR #127, opened 2018-07-20
Adds Discriminator support for Gson-OkHttp

motorro:
References: #122

Adds support for `descriminator` to Gson/OkHttp Java client using [GsonFireBuilder ](/julman99/gson-fire)1.8.3
A copy/paste of [2.0 generator](/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/resources/Java/JSON.mustache) updated to support [v3 Discriminator](/swagger-api/swagger-core/blob/master/modules/swagger-models/src/main/java/io/swagger/v3/oas/models/media/Discriminator.java)
Also supports discriminator mapping:
```yaml
    ReplyType:
      description: Reply type
      type: string
      enum:
        - accept
        - reject

    Reply:
      type: object
      required:
        - type
      discriminator:
        propertyName: type
        mapping:
          accept: AcceptReply
          reject: RejectReply
      properties:
        type:
          $ref: '#/components/schemas/ReplyType'

    AcceptReply
      allOf:
        - $ref: '#/components/schemas/Reply'
        - type: object

    RejectReply:
      allOf:
        - $ref: '#/components/schemas/Reply'
```

### DIFF
--- a/src/main/resources/v2/Java/libraries/okhttp-gson/JSON.mustache
+++ b/src/main/resources/v2/Java/libraries/okhttp-gson/JSON.mustache
@@ -4,6 +4,9 @@ package {{invokerPackage}};
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonElement;
+import io.gsonfire.GsonFireBuilder;
+import io.gsonfire.TypeSelector;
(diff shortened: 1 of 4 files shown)


HugoMario:
LGTM, thanks a lot @motorro

PR #133, opened 2018-07-26
issues-121 readme last fix

ifrolikov:


HugoMario:
pretty nice, thanks !!!

4n70w4:
Hi! @ifrolikov @HugoMario

Why line 163 of src/main/resources/v2/php/api.mustache changed from
```php
if ($returnType !== 'string') {
```
to
```php
if (!in_array($returnType, ['string','integer','bool'])) {
```

But line 259 don't changed

```php
if ($returnType !== 'string') {
```

Is that intended or a mistake?

Issue #148, opened 2018-08-09
[java] Request body's of multipart/form-data requests are ignored by at least all Java generators


lion7:
In the OpenAPI specification 3.0 an example of a multipart/form-data request body can be found ([source](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#special-considerations-for-multipart-content)):

```
requestBody:
  content:
    multipart/form-data:
      schema:
        type: object
        properties:
          id:
            type: string
            format: uuid
          address:
            # default Content-Type for objects is `application/json`
            type: object
            properties: {}
          profileImage:
            # default Content-Type for string/binary is `application/octet-stream`
            type: string
            format: binary
          children:
            # default Content-Type for arrays is based on the `inner` type (text/plain here)
            type: array
            items:
              type: string
          addresses:
            # default Content-Type for arrays is based on the `inner` type (object shown, so `application/json` in this example)
            type: array
            items:
              type: '#/components/schemas/Address'
```

The swagger-parser used by the generators parses this example different from other blocks; the schema is always a schema that refers to another schema, either an 'anonymous' schema or 1 that is defined in the components section, The current implementation of `DefaultCodegenConfig` does not follow this reference and hence all properties defined are ignored during code generation.

PR #149, opened 2018-08-09
Follow a schema reference for form content type's if a ref exists

lion7:
Fixes issue https://github.com/swagger-api/swagger-codegen-generators/issues/148


HugoMario:
thanks a lot @lion7 !!!!!

i'm going to update samples based on this.

Issue #150, opened 2018-08-09
[java] File uploads in multipart/form-data request body's are generated as a byte[] argument instead of File


lion7:
When attempting to specify file uploads in a `multipart/form-data` schema the Java generator will generate a method that takes a `byte[]` argument. But the OpenAPI Specification 3.0 states that byte arrays are properties defined as:
```
requestBody:
  content:
    multipart/form-data:
      schema:
        type: object
        properties:
          myByteArray:
            type: string
            format: byte
```
Specifying a binary type such as a file upload can be defined by using the binary format:
```
requestBody:
  content:
    multipart/form-data:
      schema:
        type: object
        properties:
          myFileUpload:
            type: string
            format: binary
```
I'd expect that in the generated Java code this would be a `java.util.File` argument. That way the used HTTP client can properly determine the `Content-Type` and `Content-Disposition` headers, where the latter contains the filename. These headers are both mandatory when submitting a file upload according to the [multipart specification](https://tools.ietf.org/html/rfc7578#section-4.2).

lion7:
Closing this as the PR has been merged.

PR #151, opened 2018-08-09
Map the binary type to a java.util.File instead of byte[]

lion7:
Fixes issue https://github.com/swagger-api/swagger-codegen-generators/issues/150

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaCodegen.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaCodegen.java
@@ -120,6 +120,7 @@ public abstract class AbstractJavaCodegen extends DefaultCodegenConfig {
         instantiationTypes.put("map", "HashMap");
         typeMapping.put("date", "Date");
         typeMapping.put("file", "File");
+        typeMapping.put("binary", "File");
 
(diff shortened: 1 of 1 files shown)


lion7:
@HugoMario mind looking at this PR?

HugoMario:
hey @lion7 !

sorry for delay response. i've been really with lack of time i'll try to take a look the PR this week.

HugoMario:
@lion7 i checked this, and looks good to me, i'm just concerned about how this could affect the output for whom expect the `byte array`.

do you think it's possible to handle this in a way we can  get any of those output according what we want ?

lion7:
I tried to explain this in the linked issue https://github.com/swagger-api/swagger-codegen-generators/issues/150. 
Basically if you want a `byte array` you would set the format to `byte` instead of `binary`. 
If you open the [OAS 3.0 specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) and search for 'multipart content' you will see that it explicitly states that the `binary` type is expected to be a file. Consider a webservice where you can upload videos, these files could easily be several gigabytes. Handling them as a `byte array` in code will most probably result in an `OutOfMemoryError` somewhere.

PR #163, opened 2018-09-04
Scala generators - Modify scala model generation

rbiersbach:
-  Profit from the already provided documentation, by generating scala doc that includes examples and more
- Remove default parameters for options that are a common source of problems and add nothing to the implementation


HugoMario:
thanks a lot @rbiersbach !!

PR #181, opened 2018-09-23
Fix path assertions so the unit tests also work on Windows

lion7:


HugoMario:
thanks a lot @lion7 !

PR #184, opened 2018-09-26
Handle the schema types 'file' and 'binary' identical

lion7:
In the Swagger 2.0 specification you could specify a file upload using the 'special' parameter type `file`.
This type does not exist in the OpenAPI 3.0 specification, instead the `binary` format should be used.

Currently (almost) all mustache/handlebars templates check a file using the expression `{{#is this 'file'}}` 
An solution to avoid changing all templates is to ensure that these 2 types are configured identically as has been done in this pull request. I will create another pull request where all relevant templates are changed to check for the `binary` type instead. 

@HugoMario could you review both PR's and pick the most preferred solution?

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
@@ -1024,7 +1024,7 @@ public abstract class DefaultCodegenConfig implements CodegenConfig {
         } else if (schema instanceof BinarySchema) {
             return SchemaTypeUtil.BINARY_FORMAT;
         } else if (schema instanceof FileSchema) {
-            return "file";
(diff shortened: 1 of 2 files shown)


HugoMario:
thanks a lot @lion7 sorry for so loooong delay. Honestly i thought this one had been alreayd merged

lion7:
@hugomario np, I'm happy that it at least got merged :smile:

lion7:
@hugomario you might also want check out https://github.com/swagger-api/swagger-codegen-generators/issues/185 and https://github.com/swagger-api/swagger-codegen-generators/issues/373 which contains a more correct fix

HugoMario:
yes, i'll take a time for them this comming week. word !

mfvitale:
Hi, 
are there fix on some release?

PR #185, opened 2018-09-26
Change expressions with 'file' type to 'binary' type

lion7:
In the Swagger 2.0 specification you could specify a file upload using the 'special' parameter type `file`.
This type does not exist in the OpenAPI 3.0 specification, instead the `binary` format should be used.

Currently (almost) all mustache/handlebars templates check a file using the expression `{{#is this 'file'}}` 
An solution is to change this to `{{#is this 'binary'}}`. This does imply that all relevant templates have to be changed. I will create another pull request where this is not necessary but that requires a change to the intepretation of type's. 

@HugoMario could you review both PR's and pick the most preferred solution?


HugoMario:
thanks for your PR's i'll take a time this friday or earlyy next week to review them a merge them

HugoMario:
@lion7 this makes sense, however some generators still works with `file` type, besides is not present on OAS3, i´ll work with your PR in order to use an option that allow us to work with both data type accordinng users selection. thanks!

mfvitale:
Hi,
in which release is this fix present?

PR #193, opened 2018-10-02
disable spring cloud library

HugoMario:


gerardbosch:
Hi @HugoMario I spent some time trying to generate the `@FeignClient` for my API and I've realized thanks to the Christophe Bornet that spring-cloud library was disabled in version 3.0.1. What was the reason for this merge? It seems that FeignClient is working in 3.0.0. I was trying with 3.0.2 and was very hard to understand why I couldn't use spring-cloud library. Is not version 3.0.2 a release version?

I am trying to understand why spring-cloud support was removed. Thanks.

HugoMario:
Hello @gerardbosch, sorry for this delay answer. We're working on generators in a way to make the output could be able to run properly. Since we couldn't have this ready for spring cloud library we temporarily disabled. 

I'm going to start working on enabling this back and let you know.

Sorry for this inconvenient.

PR #210, opened 2018-10-10
added cookie support for scala akka http server generator

rbiersbach:


HugoMario:
thanks a lot @rbiersbach!!! and sorry for this big delay

PR #213, opened 2018-10-13
Add nullable property support

cognifloyd:
Allow code generators to support the [nullable](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaNullable) property in the 3.0.0, 3.0.1, and 3.0.2 specs.

This does not actually add support to any languages; it just exposes `nullable` so the generators *can* use this standard feature.

Requires: https://github.com/swagger-api/swagger-codegen/pull/8809


cognifloyd:
I just noticed some changes from openapi-generator. Apparently I neglected to add nullable to parameter.
Relevant changes:
- OpenAPITools/openapi-generator#873
- OpenAPITools/openapi-generator#889
- OpenAPITools/openapi-generator#930

I will add a similar change to parameter as well, and reference the author in the commit message.

cognifloyd:
@HugoMario updated. Now Jenkins is happy.

PR #225, opened 2018-10-25
Fixed #224 "The generated classes for services are invalid, because a…

### LINKED ISSUE #224: [typescript-angular] Missing curly brace in the generated service
### Description
The generated classes for services are invalid, because a curly brace is missing at the end.

I found this issue on swagger-codegen project:
https://github.com/swagger-api/swagger-codegen/issues/8857
but I think it belongs here.

mtormento:
… curly brace is missing at the end"

Fixed #224.
Added missing curly brace at the end of generated services for both handlebars and mustache templates.

### DIFF
--- a/src/main/resources/handlebars/typescript-angular/api.service.mustache
+++ b/src/main/resources/handlebars/typescript-angular/api.service.mustache
@@ -337,3 +337,4 @@ export class {{classname}} {
 
 {{/operation}}
 {{/operations}}
+}
--- a/src/main/resources/mustache/typescript-angular/api.service.mustache
+++ b/src/main/resources/mustache/typescript-angular/api.service.mustache
@@ -354,3 +354,4 @@ export class {{classname}} {
 
 {{/operation}}}
 {{/operations}}
+}


src/main/resources/handlebars/typescript-angular/api.service.mustache
```diff
@@ -337,3 +337,4 @@ export class {{classname}} {
 
 {{/operation}}
 {{/operations}}
+}
```
HugoMario: i think this is not the right place to put the curly brace. it should be placed between those closing tags/elements.
```
{{/operation}}
}
{{/operations}}
```

src/main/resources/mustache/typescript-angular/api.service.mustache
```diff
@@ -354,3 +354,4 @@ export class {{classname}} {
 
 {{/operation}}}
 {{/operations}}
+}
```
HugoMario: looks like it's wrong but it's actually fine, the triple right brace in operation makes things confuse `{{/operation}}}`.
i think this could be refactored to be:
```
{{/operation}}
}
{{/operations}}
```
to avoid that confuse. wdyt? @mtormento

HugoMario:
hello @mtormento thanks a lot for this PR, in order to merge it, i made some comments about a bit changes that needs to be done.

PR #266, opened 2018-12-26
[typescript-angular] fix missing } in service files and fix install issues for windows users

MBcom:
* added missing } in typescript-angular template - see #224
* switching to generate a .tgz file for the Api bundle, this solves the link issues for windows users, because the tgz file is not linked to the projects node_modules folder - it will be unpacked
*  -> modified README for local installation instructions for windows users

* solved local folder installation issue - when installing the with `npm install --save PATH_TO_BUNDLE` the ngc compiler fails in post-install step, because some dependencies are missing in output folder
* -> modified package.json for use with tgz install


HugoMario:
thanks a lot @MBcom , sorry for this long delay :/

Issue #103, opened 2018-06-18
[Spring] @JsonTypeInfo property not set to Discriminator propertyName


Gabe-Linux:
Using Spring codegen on a schema with a discriminator property populates the `@JsonTypeInfo` annotation with the `Discriminator` class represented as a string, instead of the propertyName itself.

For example, this schema:

```yaml
Example:
      type: object
      required:
        - baseType
      discriminator:
        propertyName: baseType
      properties:
        baseType:
          type: string
        otherProp:
          type: string
```

will generate this class:

```java
/**
 * Example
 */
@Validated

@javax.annotation.Generated(value = "io.swagger.codegen.languages.java.SpringCodegen", date = "2018-06-18T16:55:52.608-04:00[America/New_York]")

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "Discriminator{propertyName&#x3D;&#x27;baseType&#x27;, mapping&#x3D;null}", visible = true )
@JsonSubTypes({
  
})

public class Example   {

  @JsonProperty("baseType")
  private String baseType = null;
  
  @JsonProperty("otherProp")
  private String otherProp = null;
  
  public Example baseType(String baseType) {
    this.baseType = baseType;
    return this;
  }
 

  /**
   * Get baseType
   * @return baseType
  **/
 
  @ApiModelProperty(required = true, value = "")
  @NotNull
  public String getBaseType() {
    return baseType;
  }

  public void setBaseType(String baseType) {
    this.baseType = baseType;
  }

  public Example otherProp(String otherProp) {
    this.otherProp = otherProp;
    return this;
  }
  
  /**
   * Get otherProp
   * @return otherProp
  **/
 
  @ApiModelProperty(value = "")
  public String getOtherProp() {
    return otherProp;
  }

  public void setOtherProp(String otherProp) {
    this.otherProp = otherProp;
  }

  @Override
  public boolean equals(java.lang.Object o) {
    if (this == o) {
      return true;
    }
    if (o == null || getClass() != o.getClass()) {
      return false;
    }
    Example example = (Example) o;
    return Objects.equals(this.baseType, example.baseType) &&
        Objects.equals(this.otherProp, example.otherProp);
  }

  @Override
  public int hashCode() {
    return Objects.hash(baseType, otherProp);
  }

  @Override
  public String toString() {
    StringBuilder sb = new StringBuilder();
    sb.append("class Example {\n");
[...]

PR #273, opened 2019-01-13
Since mustache is still default template engine for SringCodegen, app…

wvitte:
…lying fix number [103](https://github.com/swagger-api/swagger-codegen-generators/issues/103) for handlebar templates where @JsonTypeInfo property not set to Discriminator propertyName for Java Spring codegen to mustache templates as well.

Currently I see spring codegen still uses mustache as default template engine and it would be nice if this fix is applied to this template engine as long as it's used and not completely migrated to handlebars.


HugoMario:
thanks @wvitte, sorry for this delay

PR #274, opened 2019-01-14
WIP: Experimental Python support (doesn't work)

matteson:
This almost works, but I've reached a point at which I'm stuck. I'm happy to get it to work, but I need some guidance on how to proceed.

There are two big issues:
- The discriminator value to class map is not populated, and that's the last *big* issue to be resolved before I can use this for an internal project.
- The required fields of the schemas described in "oneOf" are all applied to the top level object. Specifically,

```
    foo:
      description: a foo object
      type: object
      additionalProperties: false
      discriminator:
        propertyName: class_name
      oneOf:
        - $ref: '#/components/schemas/bar-1'
        - $ref: '#/components/schemas/bar-2'
```
The required properties of `bar-1` and `bar-2` are both applied to serialization/deserialization of `foo`.

There are a few other small issues:
- There's a bad import of "Object"
- There are unused imports, that form circular dependencies for schemas defined recursively
- Tests, deprecation commit in swagger-codegen, examples are all missing
- Other things that I've probably missed.

The biggest thing I need help on is populating the discriminator value to class map. From exploring the templates, I need to set the `children` value that is passed to the mustache template. I've tried to borrow `fromModel` and `postProcessAllModels` from the Java client generator, but I'm missing something.

If there are obvious fixes for the other issues I'm having, please point me in the right direction.


HugoMario:
hello @matteson, python generators have been already added, so going to close this PR. Sorry for this delay

PR #280, opened 2019-01-20
[typescript-angular] Update default Angular version from 4.3 to 6.0

### LINKED ISSUE #279: [typescript-angular] Update default Angular version from 4.3 to 6.0
Angular 4.0.0 LTS [ended](https://angular.io/guide/releases#support-policy-and-schedule) on September 23, 2018. It should be safe now to upgrade the default Angular build version to 6.0.0, which is an already supported version since `swagger-codegen` [v2.4.0](https://github.com/swagger-api/swagger-codegen/releases/tag/v2.4.0).

Upgrading to v6 would also prevent issues like [these](https://github.com/swagger-api/swagger-codegen/issues/8856), where people who use Angular >= 6.0 can't get a working package without adding `--additional-properties ngVersion=6` to the generate script.

dennisameling:
Closes #279

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/typescript/TypeScriptAngularClientCodegen.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/typescript/TypeScriptAngularClientCodegen.java
@@ -119,7 +119,7 @@ public class TypeScriptAngularClientCodegen extends AbstractTypeScriptClientCode
         if (additionalProperties.containsKey(NG_VERSION)) {
             ngVersion = new SemVer(additionalProperties.get(NG_VERSION).toString());
         } else {
-            ngVersion = new SemVer("4.3.0");
+            ngVersion = new SemVer("6.0.0");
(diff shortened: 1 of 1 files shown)


HugoMario:
thanks a lot @dennisameling !!!

src/main/java/io/swagger/codegen/v3/generators/typescript/TypeScriptAngularClientCodegen.java
```diff
@@ -119,7 +119,7 @@ public void processOpts() {
         if (additionalProperties.containsKey(NG_VERSION)) {
             ngVersion = new SemVer(additionalProperties.get(NG_VERSION).toString());
         } else {
-            ngVersion = new SemVer("4.3.0");
+            ngVersion = new SemVer("6.0.0");
```
urban-cookie: @dennisameling Please note that the help still states that "Default is '4.3'". Probably should extract that magic string to a constant and fix that.

PR #288, opened 2019-01-30
fix anchors in docs

garipovazamat:
In doc templates several anchors was incorrect and I fixed it.


HugoMario:
thanks a lot @garipovazamat

PR #290, opened 2019-01-31
Fix swagger-codegen/issues/9126 [CSHARP] v3.x with csharp generator crashes when schema inheritence has length more than 2 

rjvdboon:
See attached a commit that causes the CSharp code generator not to crash anymore when generating code for http://developers.strava.com/swagger/swagger.json.
The generated code itself is still wrong, but I'm working on a fix for that as well.


rjvdboon:
The crash has been resolved by now in mainline, and this PR conflicts with current codebase anyway, so closing as "obsolete".

Issue #153, opened 2018-08-10
generate python-flask server stubs from openapi 3 definition


mnagel:
Hello there,

https://editor.swagger.io/ direct me here for questions about the server code generation.
My question is:
Is it correct that at the moment python-flask server stubs can be generated from swagger 2 definitions but not yet from openapi 3 definitions? If it is correct, are there any plans for the addition of this feature?

Best Regards
Michael

PS: is this the place to report general issues with the hosted instance of the swagger editor? There are some 500 server errors and some mixed content errors when using HTTPS...

HugoMario:
Hello @mnagel, yes' i hope we get this generator soon for openapi 3.

chinmayprabhudesai:
Any updates on this?

HugoMario:
@chinmayprabhudesai just added python client, need to add this one to prio tasks

matthaeusheer:
Any updates on this issue? Cheers

chinmayprabhudesai:
> Any updates on this issue? Cheers

#292 was Merged into the master branch. It There's a python client generator but I don't see a server generator

HugoMario:
sorry for delay guys, just added flask connexion with this PR #387

matthaeusheer:
When will the python-flask server generation for openapi3 be available on SwaggerHub?

anselmos:
@HugoMario It looks like you're missing : 

```java
        if (StringUtils.isBlank(templateDir)) {
            embeddedTemplateDir = templateDir = getTemplateDir();
        }
```
At PythonFlaskConnexionCodegen.java - otherwise it does not read template-dirs properly and just skips to default from jar. Related issue : https://github.com/swagger-api/swagger-codegen/issues/9363

HugoMario:
thanks @anselmos for this catch!

HugoMario:
`python-flask` generator has been added, so i think it's time to close this issue.

PR #292, opened 2019-02-01
Python generator

HugoMario:


kavdev:
Closes #153

marcinole:
@kavdev does it close #153 ? It seems to me that #153 is about generating the server stub for python-flask, not the client for python!
I'm new to swagger I've spent some time trying to figure out, so please let me know if that is correct.
Running the command : `java -jar swagger-codegen-cli.jar langs` , shows `python` in the list now but no `python-flask`

PR #298, opened 2019-02-06
When adding CodegenParameter to multiple lists, ensure it is copied f…

### LINKED ISSUE #297: x-has-more incorrectly set on CodegenOperation allParams
When using `allParams`, if there is a single `bodyParam` alongside other parameters, the vendorExtension `x-has-more` is overwritten with false. Most often, this ends up in missing commas in the generated files. 

This is due to the bodyParam (and formParam) not being set as a copy in the various lists, instead the lists point to the same object.

An example test case is adding a `put` with body and path param to `petstore-with-fake-endpoints-models-for-testing.yaml` taken from the swagger-codegen project.
```yaml
  '/pet/{petId}':
    put:
      tags:
        - pet
      summary: Updates a pet in the store with form data
      description: ''
      operationId: updatePetWithBody
      consumes:
        - application/json
      produces:
        - application/json
      parameters:
        - name: petId
          in: path
          description: ID of pet that needs to be updated
          required: true
          type: integer
          format: int64
        - name: body
          in: body
          description: Pet to update
          required: true
          schema:
            $ref: '#/definitions/Pet'
      responses:
        '405':
          description: Invalid input
      security:
        - petstore_auth:
            - 'write:pets'
            - 'read:pets'
```

samuelhewitt:
…irst.

Compiled and then tested with latest `swagger-codegen` branch `3.0.0` with the described change to the petstore...yaml.

Closes #297

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
@@ -2024,7 +2024,7 @@ public abstract class DefaultCodegenConfig implements CodegenConfig {
                             }
                             // todo: this segment is only to support the "older" template design. it should be removed once all templates are updated with the new {{#contents}} tag.
                             formParameter.getVendorExtensions().put(CodegenConstants.IS_FORM_PARAM_EXT_NAME, Boolean.TRUE);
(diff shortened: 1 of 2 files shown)


HugoMario:
thanks a lot @samuelhewitt

PR #300, opened 2019-02-09
Removed duplicated code accidentially introduced in commit 61275df

skrysmanski:
As it says it in the title.


HugoMario:
thanks @skrysmanski

PR #301, opened 2019-02-09
Fix: x-has-more not working for request body properly

skrysmanski:
This fixes the problem that the vendor extension `x-has-more` doesn't work properly for request bodies. 

The reason is that the same `bodyParam` instance is added to both the `bodyParams` and the `allParams` collection. Then, later this is done:

```java
codegenOperation.allParams = addHasMore(allParams);
codegenOperation.bodyParams = addHasMore(bodyParams);
```

The second line will set the vendor extension `x-has-more` to `false` for the last body param - even if it's not the last param in `allParams`. And that's wrong.


skrysmanski:
I'm closing this because pull request #298 is way more comprehensive and fixes the same problem.

PR #303, opened 2019-02-09
Fixed consumes list not being resolved when request body is specified via reference (rather than inline)

### LINKED ISSUE #302: Empty consumes list when request body is referenced
When the request body for an operation is defined through a `$ref` rather than inline, the "consumes" list is empty.

For example, assume this OAS3 spec:

```yaml
openapi: 3.0.0
info:
  title: Test Api
  version: '3.0.0'

paths:
  /pet:
    post:
      summary: Add a new pet to the store
      operationId: addPet
      requestBody:
        $ref: "#/components/requestBodies/Pet"
      responses:
        "405":
          description: Invalid input

components:
  requestBodies:
    Pet:
      content:
        application/json:
          schema:
            $ref: "#/components/schemas/Pet"
        application/xml:
          schema:
            $ref: "#/components/schemas/Pet"
      description: Pet object that needs to be added to the store
      required: true
  schemas:
    Pet:
      type: object
      properties:
        id:
          type: integer
          format: int64
```

The problem now is that the method `DefaultCodegenConfig.addConsumesInfo()` (called by `fromOperation()`) starts with these lines:

```java
if(operation.getRequestBody() == null || operation.getRequestBody().getContent() == null || operation.getRequestBody().getContent().isEmpty()) {
    return;
}
```

Because the request body is defined via `$ref`, the call to `operation.getRequestBody().getContent()` returns `null` and the method exits.

Possible solution: `addConsumesInfo()` should be given access to the `OpenAPI` object which contains the request body components required for resolving all the information in this method.

skrysmanski:
Fixes #302 .

The only thing I could not figure out is what should happen if the `openAPI` parameter is `null`. The current codes throws a `NullPointerException` - the same behavior as `fromOperation()`.

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
@@ -1890,7 +1890,7 @@ public abstract class DefaultCodegenConfig implements CodegenConfig {
             codegenOperation.getVendorExtensions().put(CodegenConstants.IS_DEPRECATED_EXT_NAME, operation.getDeprecated());
         }
 
-        addConsumesInfo(operation, codegenOperation);
+        addConsumesInfo(operation, codegenOperation, openAPI);
 
         if (operation.getResponses() != null && !operation.getResponses().isEmpty()) {
             ApiResponse methodResponse = findMethodResponse(operation.getResponses());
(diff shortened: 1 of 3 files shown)


skrysmanski:
Sorry about this messy history in this pull request. I was trying to rebase and squash my commits and it was not so easy to figure out how this should work. But now the pull request is fine.

HugoMario:
@skrysmanski, this looks good to me, can you please resolve conflict in order to merge your PR?

skrysmanski:
@HugoMario Done.

HugoMario:
awesome, thanks!

PR #304, opened 2019-02-10
Added and fixed some base files

skrysmanski:
Changes:

* Added `test-output` folder to Git's ignore list. This folder is generated by testng.
* Added `.editorconfig` file. This prevents people from accidentally checking in tabs instead of spaces for java files (if they have an editor with editorconfig support).


HugoMario:
awesome!!! thanks a lot @skrysmanski

PR #309, opened 2019-02-17
Fixed "getCollectionFormat()"

skrysmanski:
The current implementation of `DefaultCodegenConfig.getCollectionFormat()` is wrong. It doesn't match the behavior specified here:

https://swagger.io/docs/specification/serialization/#query

Especially:

* When `style` is `FORM` and `explode` is `true`, `csv` is returned instead of `multi`.
* When `style` is not `FORM` and `explode` is `true`, the "style value" is returned instead of `multi`. The "style value" should only be returned if `explode` is `false`.


HugoMario:
thanks a lot .  @skrysmanski !

Issue #326, opened 2019-03-12
Response header references are not resolved


chris-prop:
We have our responses set up as follows:
```
  /users/{userId}:
    get:
      description: Returns details for an individual user
      operationId: getUserById
      tags:
        - user
      parameters:
        - $ref: '#/components/parameters/userId'
        responses:
        '200':
          description: Success
          headers:
            content-language:
              $ref: '#/components/headers/contentLanguage'
            api-request-id:
              $ref: '#/components/headers/apiRequestId'
```
`DefaultCodegetConfig` not resolving the references when adding headers.

frantuma:
fixed in #327

PR #327, opened 2019-03-13
Resolve header references in responses. Added tests

chris-prop:
Fixes issue #326


HugoMario:
hi @chris-prop sorry for delay. i want to let you know that `OpenAPI` object has global access for `DefaultCodegenConfig` class. So if you would like to update your PR based on that, it could be great.

chris-prop:
@HugoMario I finally got around to updating this.

skrysmanski:
Isn‘t the propose change a breaking change (because a public method is deleted)?

chris-prop:
> Isn‘t the proposed change a breaking change (because a public method is deleted)?

The only public method I deleted was the one I added in the original PR (that was never merged).

skrysmanski:
You‘re right. My bad.

chris-prop:
Is there something I need to do to get this merged?

frantuma:
thanks!

PR #331, opened 2019-03-17
Mark form parameters as required when body is required

skrysmanski:
This pull request fixes the problem that when a form request body is marked as "required", then all form parameters should also be marked as "required".

For example, take this (shortened) input spec:

```yaml
paths:
  /test_required:
    post:
      summary: Operation with form body that is required
      requestBody:
        required: true
        content:
          application/x-www-form-urlencoded:
            schema:
              $ref: '#/components/schemas/Category'

components:
  schemas:
    Category:
      type: object
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
```

It results in 2 form parameters (`id` and `name`). Without this fix both parameters would be marked as "optional", even though the request body is marked as "required".


HugoMario:
awesome, thanks !!!

Issue #350, opened 2019-04-03
[java]- Client sends empty Accept header


jweisman:
If the response contains no content tag (i.e. `DELETE` requests), the Java client sends an empty `Accept` header, which causes problems in some servers (specifically, 406 Not Acceptable errors).

Client should not send an Accept header if no content type is available in the spec.

PR #351, opened 2019-04-03
[java] Client should only send accept header if not null

jweisman:
Only send `Accept` header if not null. Resolves swagger-api/swagger-codegen-generators#350


HugoMario:
thanks a lot @jweisman

Issue #353, opened 2019-04-03
JavaSpring: Compilation error when using path parameters with enum values


npoirier:
### Issue description

Using swagger-codegen-maven-plugin 3.0.6 or 3.0.7, API interfaces generated for JavaSpring contain extra quotes around allowable values for PathVariable parameters, which  causes compilation errors.

```java
@ApiOperation(value = "", nickname = "getTest", notes = "", tags={ "test-controller", })
    @ApiResponses(value = { 
        @ApiResponse(code = 200, message = "OK") })
    @RequestMapping(value = "/foo/{enumValue}",
        method = RequestMethod.GET)
    ResponseEntity<Void> getTest(@ApiParam(value = "",required=true, allowableValues = ""value1", "value2"") @PathVariable("enumValue") String enumValue);
```

`allowableValues` was not generated previously with swagger-codegen-maven-plugin 3.0.5.

### Steps to reproduce

Generate a client using swagger-codegen-maven-plugin 3.0.6 or 3.0.7 with a Swagger spec containing a Path parameter with an enum value:

```---
swagger: '2.0'
info:
  title: test-service
basePath: /
tags:
  - name: test-controller
    description: Test Controller
schemes:
  - http
paths:
  /foo/{enumValue}:
    get:
      tags:
        - test-controller
      parameters:
        - in: path
          name: enumValue
          type: string
          enum:
            - value1
            - value2
      operationId: getTest
      responses:
        '200':
          description: OK
```

PR #354, opened 2019-04-04
issue-353 - JavaSpring - Fix allowedValues for enums in path parameters

npoirier:
Output allowedValues for path parameters the same way its done for header and query params.

Resolves https://github.com/swagger-api/swagger-codegen-generators/issues/353


HugoMario:
thanks a lot @npoirier

PR #355, opened 2019-04-05
Fix generated python api imports to include packageName.

oleastre:
Generate base python package import does not work at all (tested with python 3.7).
This PR fixes the generated importPath to correctly include the base package name.


HugoMario:
thanks @oleastre

PR #356, opened 2019-04-08
Typescript import must use / as path separator in every operating system

dbusacca:
Typescript import must use / as path separator in every operating system.

Using File.separator in Windows generates invalid typescript code:

### Before:
`import { MailService } from './api\mail.service';`  <- Invalid

### After:
`import { MailService } from './api/mail.service';`  <- correct in every operating system


hypery2k:
this PR is obsolete, already in master

PR #362, opened 2019-04-15
Added escaping of JSON properties with special char @ (e.g. in "@id" and "@type" of NuGet REST API)

thomas-worm-datev:
Some REST API contain JSON object property names containing the special char @. E.g. the NuGet REST API contains "@id" and "@type". This fixed compilation errors with such properties.


HugoMario:
thanks a lot @thomas-worm-datev !!

PR #363, opened 2019-04-16
Use correct retrofit annotation for query parameters

hbakkum:
Fixes an issue when using the mustache templates for generating classes using retrofit2 library. Query parameters get marked with the retrofit Path annotation which breaks any service endpoints using query parameters. This was hit when using the swagger-codegen-maven-plugin v3.0.7.


HugoMario:
nice catch, thanks!!

PR #364, opened 2019-04-16
issue-260 - Fix boolean property double 'is' issue in Java codegen.

dalbrekt:
'is' is already prepended in https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaCodegen.java#L1342


HugoMario:
thanks a lor @dalbrekt !!!!!

PR #372, opened 2019-04-28
fixed incorrect path for imports

HugoMario:
fixes https://github.com/swagger-api/swagger-codegen/issues/8930


paulbors:
Thanks guys for this, when in the next release scheduled?

PR #377, opened 2019-05-06
Fix CodegenModel.allVars to include vars of composed and parent models

nikhilym:
## Description
The `allVars` field in `CodegenModel` is doesn't contain the composed model properties and only contains the inherited properties. This is because the `allProperties` field passed during [`addVars`](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java#L1382) in ComposedSchema instance check of `fromModel` doesn't contain the properties of the composed model.

This commit adds the code to add the composed model properties to `allProperties` variable, that is used to add the variables in `allVars` variable. This is similar to how it is handled in the 2.0 [DefaultCodegen.java](https://github.com/swagger-api/swagger-codegen/blob/master/modules/swagger-codegen/src/main/java/io/swagger/codegen/DefaultCodegen.java#L1464) file.

## Testing
- Used the following petstore yaml file for checking :
````
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  description: A sample API that uses a petstore as an example to demonstrate features in the OpenAPI 3.0 specification
  termsOfService: http://swagger.io/terms/
  contact:
    name: Swagger API Team
    email: apiteam@swagger.io
    url: http://swagger.io
  license:
    name: Apache 2.0
    url: https://www.apache.org/licenses/LICENSE-2.0.html
servers:
  - url: http://petstore.swagger.io/api
paths:
  /pets:
    get:
      description: |
        Returns all pets from the system that the user has access to
      operationId: findPets
      parameters:
        - name: tags
          in: query
          description: tags to filter by
          required: false
          style: form
          schema:
            type: array
            items:
              type: string
        - name: limit
          in: query
          description: maximum number of results to return
          required: false
          schema:
            type: integer
            format: int32
      responses:
        '200':
          description: pet response
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Pet'
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Error'
components:
  schemas:
    Pet:
      allOf:
        - $ref: '#/components/schemas/NewPet'
        - required:
          - id
          properties:
            id:
              type: integer
              format: int64
    NewPet:
      required:
        - name
      properties:
        name:
          type: string
        tag:
          type: string
    Error:
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
````
- Changed the python [`model.mustache`](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/resources/handlebars/python/model.mustache#L39) handlebar to use `allVars` in `swagger_types`.
- Following is the generated output for `Pet` class before the changes (pasting only the swagger_types part)
````
class Pet(NewPet):
    """NOTE: This class is auto generated by the swagger code generator program.

    Do not edit the class manually.
    """
    """
    Attributes:
      swagger_types (dict): The key is attribute name
                            and the value is attribute type.
      attribute_map (dict): The key is attribute name
                            and the value is json key in definition.
    """
    swagger_types = {
        'name': 'str',
        'tag': 'str'
    }
````
- Following is the generated output for `Pet` class after the changes
````
class Pet(NewPet):
    """NOTE: This class is auto generated by the swagger code generator program.

    Do not edit the class manually.
    """
    """
    Attributes:
      swagger_types (dict): The key is attribute name
                            and the value is attribute type.
      attribute_map (dict): The key is attribute name
                            and the value is json key in definition.
    """
    swagger_types = {
        'name': 'str',
        'tag': 'str',
        'id': 'int'
    }
````
- `mvn surefire:test` passes.
## Environment
- **Swagger-codegen-generator** version : 1.0.9-SNAPSHOT 
- **Swagger-codegen** version : `3.0.0` branch, 3.0.9-SNAPSHOT


HugoMario:
thanks a lot @nikhilym !!!
pretty helpful PR

nikhilym:
Thanks @HugoMario. Glad it helped 😄

PR #380, opened 2019-05-09
Set the json model to null in the OneOf and AnyOf interfaces

oleastre:
The OneOf and AnyOf interfaces added as part of PR #367 broke the html doc generation.
The interfaces are added as models in the javascript generation with:
```
defs.{{name}} = {{{modelJson}}};
```
Since `modelJson` was not set for those interfaces, the generated javascript is invalid.
This commit quickly fixes the problem by setting `modelJson` to the json `null` value. 
This is sufficient to make the html2 template happy and it's the only use of the `modelJson` field, so probably no effect on other templates.


HugoMario:
hi @oleastre, thanks for catching the issue and create this PR, it seems this fix the problem on HTML generators but it will break the feature for others generators as Java. is it possible you manage these changes that only affects the HTML generators?

i would recommend calling the SchemaHandler class in html generators and make changes there. you could add new method, new extended class overriding the methods or something similar that affect only to HTML generators.

please , let me know if you have any doubt about it

oleastre:
I used that way because the `modelJson` fields of the `CodegenModel` does not seems to be used in other templates. I don't see how it could break the java templates. 
When I search for `modelJson`, I only find the following 3 references (without my changes):
```
src/main/resources/mustache/htmlDocs2/index.mustache:            defs.{{name}} = {{{modelJson}}};
src/main/resources/handlebars/htmlDocs2/index.mustache:            defs.{{name}} = {{{modelJson}}};
src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java:        codegenModel.modelJson = Json.pretty(schema)
```
But it's maybe used in another way I haven't found. My goal is to do the correct thing without breaking other templates, not only for the html2 generator (in fact I also generate python client code for my current project).
If you have better ideas, I would be happy to implement it.
Also, how can I test it my changes do not break the java generator ?

PR #381, opened 2019-05-14
Remove duplicate "is" from Boolean getter

### LINKED ISSUE #260: [java] Duplicate "is" in boolean getter method names
Here an example of generated boolean field

```java
 /**
   * Tells if the resource is enabled
   * @return enabled
  **/
  @JsonProperty("enabled")
  public Boolean isisEnabled() {  // <<<< Should be isEnabled
    return enabled;
  }

  public void setEnabled(Boolean enabled) {
    this.enabled = enabled;
  }

  public Entity enabled(Boolean enabled) {
    this.enabled = enabled;
    return this;
  }
```

jweisman:
Resolved #260


HugoMario:
thanks @jweisman, going to check the problem with jenkins in order to merge this PR

PR #385, opened 2019-05-21
Add petstore.yaml and test its generation in AkkaHttpServerCodegenTest

karolchmist:
I'm working on `AkkaHttpServerCodegen` and I wanted to test quickly some changes. I've added this simple test to make sure I don't break the generation of the files.

It was inspired by this test:
https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/test/java/io/swagger/codegen/v3/generators/java/JavaModelTest.java#L1019


HugoMario:
thanks a lot @karolchmist  !!!

PR #386, opened 2019-05-21
Replace dataType with baseType+containerType to fix compilation error in AkkaHttpServerCodegen

karolchmist:
Current version of `scala-akka-http-server` generator doesn't produce correct code if the response type is `array`. For example, for petstore it produces 

```scala
def findPetsByStatus200(responseList[Pet]: List[Pet])
(implicit toEntityMarshallerList[Pet]: ToEntityMarshaller[List[Pet]]): Route =
   complete((200, responseList[Pet]))
```

This code is not valid because `responseList[Pet]` etc contains the type with brackets (`{{dataType}}` in `api.mustache` templates).

This PR fixes this by using `{{baseType}}{{containerType}}` instead, which produces:
```scala
def findPetsByStatus200(responsePetarray: List[Pet])
(implicit toEntityMarshallerPetarray: ToEntityMarshaller[List[Pet]]): Route =  
   complete((200, responsePetarray))
```

I had to change the type of `complexReturnTypes` from `Set<String>` to `List<CodegenResponse>`, which is necessary as far as I understand the code.

I'm not a Java developer, and it's the first time I use this project, so please let me know if something can be improved.


HugoMario:
merged!!! thanks again @karolchmist !

PR #393, opened 2019-06-13
Swift 4 Updates

apfritts:
- Resolve some compiler warnings
- Resolve error when 2 enums are present on a model
- Allow a more lenient / newer version of Alamofire in Podspec


apfritts:
I'm not going to have time to continue this work. Anyone is welcome to take it over.

Issue #395, opened 2019-06-18
tox python version is incoherent with docker image


ioggstream:
## I expect

tox python version to be coherent with the one in Dockerfile

## instead 

it's not:

- tox uses py35
- Dockergile uses python:3-alpine referencing .37

PR #396, opened 2019-06-18
Fix: #395. Use a coherent python version.

ioggstream:
## This PR

Fixes python version making it coherent between  Dockerfile and tox moustaches.


HugoMario:
thanks @ioggstream

PR #406, opened 2019-06-27
fix unit test cases with wrong expect path

qiuminwei:
The path to the templateDir is spliced by backslash with hardcode in
DefaultCodegenConfig, so these test cases here will fail on Windows.


HugoMario:
hello @qiuminwei , thanks God, you didnt close this too xD

i'll take a look on it after release (tomorrow) and will help you with merge.

forker-42:
Hello,

I made a similar PR where I take the opposite way to fix it. I think `File.separator` should be used to favor platform-independent code.

PR #415, opened 2019-07-10
refs swagger-api/swagger-codegen#9548 - fix NPE for composed models

frantuma:


chrisinmtown:
will this be released as 3.0.10 or something else?

frantuma:
yes, it will be part of upcoming 3.0.10

PR #425, opened 2019-07-18
Fixe python model template to work with inheritance

giavjeko:
There are several issues with the python model template that causes incorrect behavior with inherited models.

1. Subclass constructor never calls the super constructor or even takes the super constructor arguments.
2. Subclass `swagger_types` and `attribute_map` don't include parent's properties. Because of that, all parent class properties will be removed in `ApiClient.__deserialize_model`.
3. OpenAPI specification v3 supports discriminator mapping by custom names ([docs](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/)). That allows us to have custom strings in discriminator fields which then need to be mapped to class names. Currently, the model template does not render those mappings correctly.

This change fixes all the issues above.


HugoMario:
great! thanks a lot @giavjeko !!

PR #426, opened 2019-07-19
Fix python template dir path concatenation

giavjeko:
If custom python template dir is provided it gets discarded on path concatenation, ending up with a relative path based on the current working directory.

This change fixes the issue.


giavjeko:
This was actually wrong, closed

PR #427, opened 2019-07-22
fix duplicate key in generated json when JsonTypeInfo is used

bsa01:


HugoMario:
LGTM, thanks a lot @bsa01 !!

PR #428, opened 2019-07-22
Fix Python client custom template path

giavjeko:
If set through CLI using `--template-dir` option, the final path where the generator was looking for templates was non-existing.

This fixes the issue using the same approach as in [JavaClientCodegen](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/v3/generators/java/JavaClientCodegen.java#L123).


anselmos:
This has been fixed at [this commit](https://github.com/swagger-api/swagger-codegen-generators/commit/55863bc181165e1e98d74dd41f33deb3fc0cabaa) @giavjeko

frantuma:
Thanks for the PR, as mentioned by @anselmos this is included and replaced by #445

PR #431, opened 2019-07-24
Fix recursion failure when generating examples

scottydawg:
This fixes an issue documented in [Issue 9546 in the swagger-codegen repo](https://github.com/swagger-api/swagger-codegen/issues/9546). The ExampleGenerator recurses infinitely when a schema object is self-referencing. 

This fixes the issue by indexing the processed models by their schema reference, and returning existing examples when they already exist.

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/examples/ExampleGenerator.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/examples/ExampleGenerator.java
@@ -157,7 +157,10 @@ public class ExampleGenerator {
     }
 
     private Object resolveSchemaToExample(String propertyName, String mediaType, Schema schema, Set<String> processedModels) {
-//        logger.debug("Resolving example for property {}...", schema);
+        if (processedModels.contains(schema.get$ref())) {
+            return schema.getExample();
+        }
+        processedModels.add(schema.get$ref());
         if (schema.getExample() != null) {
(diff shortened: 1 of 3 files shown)


src/main/java/io/swagger/codegen/v3/generators/examples/ExampleGenerator.java
```diff
@@ -269,10 +272,10 @@ private double randomNumber(Double min, Double max) {
     }
 
     private Object resolveModelToExample(String name, String mediaType, Schema schema, Set<String> processedModels) {
-        if (processedModels.contains(name)) {
+        if (processedModels.contains(schema.get$ref())) {
             return schema.getExample();
         }
-        processedModels.add(name);
+        processedModels.add(schema.get$ref());
```
scottydawg: These `processedModels` are passed to `resolveSchemaToExample()` at line 290.

I changed this from using the `name` because names are not guaranteed to be unique within an object tree. For example, an `id` property on the top-level schema could be redundant with an `id` in a nested object. In the case of a self-referencing or circular structure, names are certain _not_ to be unique.

HugoMario:
hello @scottydawg, thanks a lot for PR, i reviewed and it does fix the stackoverflow issue, but add another one, usually for petstore sample in aspnet generated code we use to get an example json of the schemas, however with your changes we're getting this output:

```
exampleJson = "[ null, null ]";
```
i'll try to work with your chages and improve the class in order to merge the PR, thanks!

PR #433, opened 2019-07-25
Changes to use discriminator properties if specified in client code generation

nitinjavakid:
Earlier the discriminator mapping values were ignored. Now specified discriminator value is correctly used.

Also, earlier discriminator value was not set in request by client. This fixes that too.


HugoMario:
hey @nitinjavakid, i'll add a sample based on your changes. Thanks a lot for this PR!

PR #439, opened 2019-08-04
added bearer authentication description on documentation templates an…

HugoMario:
…d capability to use a sample token on operations


HugoMario:
related to https://github.com/swagger-api/swagger-codegen/issues/9365

PR #454, opened 2019-08-30
Added Java Spring Boot v2 language

ferraro:
As Spring Boot v1 is in maintenance mode only, I cloned JavaSpring to JavaSpring2 and updated the code to be compatible with Spring Boot v2.
I would not yet delete the code for v1, as maybe API providers still need it to not break their implementations.


HugoMario:
hello @ferraro, 

According to this PR; you're adding a new generator
```
https://github.com/swagger-api/swagger-codegen-generators/pull/454/files#diff-dc2e1a46dd60662b8af87c5255162b1cR18
```
but the java class named Spring2Codegen is not included in PR

ferraro:
I just checked the PR, I see the Spring2Codegen class, see screenshot:
Screenshot 2019-09-04 at 13 09 01

HugoMario:
oh sorry, my bad, thanks a lot for this PR, merging now!!!

PR #456, opened 2019-08-30
fix for Bug generating Boolean type property

gracekarina:
fix for #9203 jaxRS-cxf-client


HugoMario:
thanks a lot @gracekarina !!!

Issue #467, opened 2019-09-16
Default value for optional init params in Swift


deanWombourne:
Hi.

### Summary:

It would be really nice if Swift model objects had a default nil value for optional parameters so for model objects with many optional parameters we can create them more simply. There's a similar issue on the swagger-codegen repo https://github.com/swagger-api/swagger-codegen/issues/8365. This is a trivial, backwards compatible change to the swift handlebars template.

### Details:

Currently the template generates code like this:

```swift
public struct Order: Codable {

    public var _id: Int64
    public var petId: Int64?

    public init(_id: Int64, petId: Int64?) { 
        self._id = _id
        self.petId = petId
    }
}
```

It would be really nice if it have default values to the optional init methods, like this:

```swift
public init(_id: Int64, petId: Int64? = nil) { 
```

This seems a trivial change in the template so any objections to a pull request?

HugoMario:
Hey @deanWombourne i'm not familiar with swift4 so i cant help a lot with this feature, however if you are willing to propose a PR, i could help with merging.

PR #468, opened 2019-09-16
Fix issue 467: Add = nil to the init template in swift 4

deanWombourne:
This should fix issue #467 by adding `= nil` when the init parameter is optional.


HugoMario:
thanks a lot @deanWombourne for this PR!!

PR #471, opened 2019-09-16
Issue 470 - swift 4 multiple enum error

### LINKED ISSUE #470: [swift 4] compiler error with multiple enums in model
If we have a model with multiple enums, we generate this code:

```swift
struct MyModel: Codable {

    public enum Recipient: String, Codable { 
        case customer = "Customer"
        case shop = "Shop"
    }   public enum Channel: String, Codable {
        case formattedEmail = "FormattedEmail"
        case blankLetter = "BlankLetter"
    }
...
```

which fails compilation with the error "Consecutive declarations on a line must be separated by ';'".

This is easy to fix by adding a line break at the end of each generated enum.

deanWombourne:
Just add a line break in the model template to split out enums into their own lines.

Fixes #470 .


HugoMario:
thanks a lot @deanWombourne for this PR

PR #479, opened 2019-09-27
Kotlin-client template improvements. Less boilerplate code will be ge…

crazyk2:
…nerated


HugoMario:
thanks @crazyk2 for PR!!

PR #485, opened 2019-10-02
validate ref schema as objects.

HugoMario:


src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
```diff
@@ -4133,16 +4133,23 @@ else if (Parameter.StyleEnum.SPACEDELIMITED.equals(parameter.getStyle())) {
...
             return true;
         }
-        if (SchemaTypeUtil.OBJECT_TYPE.equals(schema.getType()) && !(schema instanceof MapSchema)) {
+        if (SchemaTypeUtil.OBJECT_TYPE.equalsIgnoreCase(schema.getType()) && !(schema instanceof MapSchema)) {
             return true;
         }
         if (schema.getType() == null && schema.getProperties() != null && !schema.getProperties().isEmpty()) {
             return true;
         }
+        if (StringUtils.isNoneEmpty(schema.get$ref())) {
```
frantuma: should be I guess `StringUtils.isNotBlank(..`

HugoMario: right, was a kind of typo

PR #491, opened 2019-10-07
Fix dateLibrary=java8 handling in JacksonConfig.mustache #9753

user667:
Previously config option dateLibrary=java8 was ignored and joda
was added instead, causing compilation failure.

Fix #9753


user667:
Fixes issue reported in project ```swagger-codegen```, see https://github.com/swagger-api/swagger-codegen/issues/9753

PR #518, opened 2019-10-29
Enum naming

### LINKED ISSUE #514: [BUG][Kotlin] Enum var naming
By default and according Kotlin naming convention, Enum variables are UPPERCASE. But swagger by default has [camelCase](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/v3/generators/kotlin/AbstractKotlinCodegen.java#L36). Why?
When I use String variable 1 with value "1", swagger generate broken code.
```Kotlin
enum class Choices1X2(val value: kotlin.String){
    1("1"),
    2("2"),
    x("X");
}
```
If look at code [sanitizeKotlinSpecificNames](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/v3/generators/kotlin/AbstractKotlinCodegen.java#L537) first add underscore ("_") before digit, but after in switch `camelCase` [camelize](https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/v3/generators/kotlin/AbstractKotlinCodegen.java#L429) delete all underscores and broken code is ready to compile. At all [ENUM_PROPERTY_NAMING_TYPE](https://github.com/swagger-api/swagger-codegen/blob/20d863d532998324d1a2d80d9d3e890b2d3184ed/modules/swagger-codegen/src/main/java/io/swagger/codegen/CodegenConstants.java#L161) only used by Kotlin generator. Why this Enum exist?
If I use UPPERCASE all works well.
So questions:
1) Kotlin allow use almost all symbols (except :;></\, for example valid variable name `value!@#$%^&*()_-+=`, valid class name `DrawSortField!@#$%^&*()_-+=`) and digits with qoute `` in Class name, Enum variable name, Class variable name. Why not use this feature instead underscore?
2) Why only for Kotlin we have so much naming starategeis for Enum? I prefer delete all bolepraite

crazyk2:
fix #514


crazyk2:
@frantuma, @webron, @HugoMario, pls take a look

HugoMario:
Hi @crazyk2,

I checked these changes locally generating code for `kotlin-client`, but it seems the `toEnumVarName` method is never called, so enums are not turned into uppercase. So having a schema like:
```yaml
components:
  schemas:
    Order:
      type: object
      properties:
        id:
          type: integer
          format: int64
        petId:
          type: integer
          format: int64
        quantity:
          type: integer
          format: int32
        shipDate:
          type: string
          format: date-time
        status:
          type: string
          description: Order Status
          enum:
            - placed
            - approved
            - delivered
        complete:
          type: boolean
          default: false
```

we got:
```java
data class Order (

    val id: kotlin.Long? = null,
    val petId: kotlin.Long? = null,
    val quantity: kotlin.Int? = null,
    val shipDate: java.time.LocalDateTime? = null,
    /* Order Status */
    val status: Order.Status? = null,
    val complete: kotlin.Boolean? = null
) {
    /**
    * Order Status
    * Values: placed,approved,delivered
    */
    enum class Status(val value: kotlin.String){
        placed("placed"),
        approved("approved"),
        delivered("delivered");
    }
}
```

PR #519, opened 2019-10-29
 Kotlin model name generation

### LINKED ISSUE #515: [BUG][Kotlin] Model class naming
`toModelName` in `AbstractKotlinCodegen` use `isReservedWord` and escape Kotlin reserved words. But for class\data class\enum class name you shouldn't use escape. All reserved words (except `Companion`) could be used as  class\data class\enum class name. `Companion` can't be used as Enum class name.

When I used word `Package` for my object class, swagger generated broken class: `data class &#x60;package&#x60; `

Need to be deleted:
```

```

crazyk2:
fix #515


crazyk2:
@frantuma, @webron, @HugoMario, pls take a look

Issue #534, opened 2019-11-11
[C# Client] JWT Authorization code is not generated for OpenApi 3 client


gavriln:
While implementing JWT Bearer Authentication per https://swagger.io/docs/specification/authentication/bearer-authentication/
```json
"security": [
  {
	"bearerAuth": [ ]
  }
]

"securitySchemes": {
  "bearerAuth": {
	"type": "http",
	"scheme": "bearer",
	"bearerFormat": "JWT"
  }
}
``` 
todoexample.json is attached as .txt
[todoexample.txt](https://github.com/swagger-api/swagger-codegen-generators/files/3832950/todoexample.txt)

I generated C# client expecting to get Authorization header to be set and wasn't seeing it.
After digging into it it turned out that mustache and handlebar templates did not have a stub for isBearer
Generated TodoApi.cs only had below comment for Authentication without any code that adds actual Authorization http header:
```
// authentication (bearerAuth) required
```

Ill be submitting a PR shortly.  Just wanted to file this issue to attach it to the PR

Jar used was built off of branch 3.0.0 https://github.com/swagger-api/swagger-codegen/tree/3.0.0

Command used to build the C# client:
```
java -jar swagger-codegen-cli.jar generate -i todoexample.json -l csharp -o csharp
```

PR #535, opened 2019-11-11
Fix for issue #534. Adding Authorization logic in the C# api temlate …

gavriln:
…class for JWT Bearer authentication.


frantuma:
Thanks! included and replaced by #557

PR #562, opened 2019-12-13
refs swagger-api/swagger-codegen#9909 - fix javascript generator templateDir and missing template file

frantuma:


Niko9911:
@frantuma Do you think there will be soon release containing this patch?

frantuma:
no ETA yet, but within middle January

PR #563, opened 2019-12-18
[AbstractJavaCodegen] Handle colons in property names

michaeldavis-wf:
Currently, properties that contain colons don't properly sanitize those values when generating POJO variable names. This PR handles those values similarly to other special characters.

Previously, POJOs were generated as such, which does not compile:

```java
@JsonProperty("user:name")
private String user:name = null;
```

Now POJOs are generated:

```java
@JsonProperty("user:name")
private String userName = null;
```


sullivanmatt:
@HugoMario if I could also request your :eyes: on this and https://github.com/swagger-api/swagger-codegen-generators/pull/584, it would be much appreciated.  Thank you!

PR #564, opened 2019-12-18
[JavaJaxRS] Fix parameter annotation generation

michaeldavis-wf:
When generating parameter annotations with the JavaJaxRS generator, those annotations are invalid and cause compilation errors. Currently, any non-cookie parameter is generated as such:

```java
public Response listUsers(@Parameter(ParameterIn.HEADER, description = "Http Authorization Header" ,required=true) @HeaderParam("Authorization"))
```

Which throws a compilation error: 

```
annotation values must be of the form 'name=value'
```

This appears to be a bug [introduced recently.](https://github.com/swagger-api/swagger-codegen-generators/pull/549)

This now generates as such, which compiles correctly: 

```java
public Response listUsers(@Parameter(in = ParameterIn.HEADER, description = "Http Authorization Header" ,required=true) @HeaderParam("Authorization"))
```


HugoMario:
thanks a lot @michaeldavis-wf!!! for this PR

Issue #592, opened 2020-02-04
Automatically request OAuth2 access token for outgoing HTTP call from client


trohrberg:
Recently, I was using OpenAPITools/openapi-generator for generating a Java/okHTTP based client SDK for my REST interface specified in OpenAPI version 3.0.2. Unfortunately, that generator had problems with inheritance defined in my models. For that - and other - reasons, I would like to us the swagger-codegen generator instead. However, that seems to miss the very handy feature of automatically requesting an access token for outgoing HTTP calls from the client SDK.

In the OpenAPITools/openapi-generator that feature is done by registering an interceptor extending the OAuth class to the okHTTP client instantiated in the ApiClient class. With that any outgoing HTTP call from the client SDK is intercepted and preceeding call to the Auth-Server is done first to retrieve an access token. That access token is then added to the original outgoing HTTP call providing an authorized access to the actual API.

Since I already figured how it works in the OpenAPITools/openapi-generator, I am going to try to provide a pull request with that feature shortly.

PR #593, opened 2020-02-04
#592 Ported automatic access token retrieval from openapi-generator

trohrberg:
Fix issue #592 

The OpenAPITools/openapi-generator generates Java client SDK with okHttp which provides an automatic retrieval of an OAuth2 access token. This is done by registering an interceptor (RetryingOAuth class) extending the OAuth class to the okHTTP client instantiated in the ApiClient class. With that any outgoing HTTP call from the client SDK is intercepted and preceeding call to the Auth-Server is done first to retrieve an access token. That access token is then added to the original outgoing HTTP call providing an authorized access to the actual API.


trohrberg:
@HugoMario - Hints for code review:

I'm not sure if the OAuthOkHttpClient.mustache and RetryingOAuth.mustache should only be included in the generating process within JavaClientCodegen.java if OAuth is specified in the corresponding YAML. In OpenAPITools/openapi-generator, from which I ported the corresponding code, it is done that way. However, the OAuth.mustache template is also always included. For that I followed the same approach for the new templates.

Issue #602, opened 2020-02-15
Generated code does not implement model inheritance correctly


trohrberg:
In our business project based on Spring Boot, we're using swagger-codegen to generate model classes and REST controller interfaces for our API described in an OpenAPI 3.0 YAML file. However, the recent release 1.0.16 does not correctly generate model classes if they come with an inheritance structure. There are basically two errors involved:

- The API model is not correctly generated into a classes and interfaces inheritance hierarchy. In our situation, a parent model which should be generated as an interface is in fact generated as a class, but the inheriting model class still uses "implements" instead of "extends" which would then be the correct statement.

- If using a discriminator mapping, the generated code does not correctly implement that due to an error in the typeInfoAnnotation.mustache template.

Since I already figured where the errors occur within the swagger-codegen-generators project, I am going to provide a pull request which fixes this issue shortly.

PR #603, opened 2020-02-15
#602 Generated code does not implement model inheritance correctly

trohrberg:
Fix handling of composed models and adjustments to model.mustache and typeInfoAnnotation.mustache templates to correctly generate models with inheritance structure.

The problem was due to three different reasons:
- The SpringCodegen source code was missing the handling of composed models.
- The model.mustache template was referring to a deprecated variable concerning composed models.
- The typeInfoAnnotation.mustache template was missing the handling of discriminator mapping.


HugoMario:
thanks a lot @trohrberg !!

Issue #621, opened 2020-03-10
[PHP] Values with a data type format of date always render as an DateTime::ATOM date-time


jpswade:
Values with a data type format of date always render as an DateTime::ATOM date-time.

For example:
```
{
          "name" : "End",
          "in" : "query",
          "description" : "Only return appointments where the `end` field is on or before the ISO8601 date value provided",
          "required" : false,
          "type" : "string",
          "format" : "date"
        }
```

This will render as "2020-03-10T19:58:40 00:00" in the format of `DateTime::ATOM`, when it's expected to be "Y-m-d".

jpswade:
Looking at the code:

`public static function sanitizeForSerialization($data, $type = null, $format = null)`

$type is never used in that function, and nothing ever passes $format in to it.

See https://github.com/swagger-api/swagger-codegen-generators/blob/789fc8aaaf61daa212797132112cc32419ab1a9f/src/main/resources/mustache/php/ObjectSerializer.mustache#L41

Or https://github.com/swagger-api/swagger-codegen-generators/blame/2cdbb459746b36d88f297268650d9c7effed3e20/src/main/resources/handlebars/php/ObjectSerializer.mustache#L41

Perhaps @HugoMario or @ackintosh could explain the intention here? Thanks.

HugoMario:
hi @jpswade, i'm not familiar with php code on templates, if it's causing the issue and you have a solution, let me know and i'll help you with PR

jpswade:
Hi @HugoMario,

I understand the problem, but I don't understand why the format isn't being passed to the method.

I'll take a look into it, but it might be easier to ask the original author what their intention was.

Do you know who originally contributed?

Cheers.

HugoMario:
hey @jpswade sorry for delay. I ignore that too, i guess you can track commits from [swagger-codegen](https://github.com/swagger-api/swagger-codegen) and figure that out.

PR #624, opened 2020-03-11
Pass Data Format through to the method

jpswade:
We now pass the data format to the toQueryValue function so we can correctly handle the data format.


jpswade:
See: https://github.com/swagger-api/swagger-codegen-generators/issues/621

HugoMario:
thanks a lot @jpswade

PR #632, opened 2020-03-18
Fixing filename concatenation issue in generation [Windows/Java JAX-RS]

### LINKED ISSUE #626: Filenames incorrect on windows due to inconsistent use of "/" and File.separator [Windows only, Jersey JAX-RS]
As of 1.0.17, becuase the [DefaultCodegenConfig#apiFilename](https://github.com/swagger-api/swagger-codegen-generators/blob/v1.0.17/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java#L3311) method uses `File.separator` to concatenate the filename, an issue occurs on windows that causes the generator to create files with incorrect package structures. This is a problem because some classes are expected to be in the same package, and this could also break ignore rules, leading to duplicate classes and other compilation issues. In the [java JAX-RS generator](https://github.com/swagger-api/swagger-codegen-generators/blob/v1.0.17/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaJAXRSServerCodegen.java#L227), we check for a `'/'`, causing the index to be inaccurate on Windows (or any other OS that doesn't use `/` as the file path separator), resulting in a the substring and replace calls not doing what is intended. The same happens in the `Factory.mustache` branch of the if statement.

We should use `File.separator` everywhere we are forming/checking file paths. Otherwise, we run the risk of assuming the OS the project is running on, and breaking compilation due to malformed paths.

will-thurman:
Fixing an issue on windows caused by inconsistent use of the OS file separator character and hardcoded `/` in the Java JAX-RS generator. Added tests for affected workflows as well.

resolves #626

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaJAXRSServerCodegen.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaJAXRSServerCodegen.java
@@ -16,6 +16,7 @@ import io.swagger.v3.oas.models.PathItem;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


will-thurman:
@HugoMario, any thoughts on this?

PR #633, opened 2020-03-20
Fixing some scala akka server http issues

Bouillie:
This PR does the following:
- the --invoker-package additional property isn't overwritten anymore.
- it fixes a bug where path parameter type were not correctly computed, resulting in a bad file generation. The problem occured when parameters were defined by reference instead of directly in the operation.
- The default value of optional query parameters is now used. Thus, the signature of generated operation function doesn't take optional parameters anymore, as the default value is used.
- it fixes a problem where multiple occurences of the same Unmarshaller were generated.


Bouillie:
I noticed a design flaw in the commit dealing with optional parameters, I'll fix it and reopen the submission later.

HugoMario:
thanks @Bouillie, let me know when PR is ready and i'll help you with merging

PR #639, opened 2020-04-02
fix: js supporting files

### LINKED ISSUE #638: [javascript] .babelrc is not valid json and babel is not able to compile tests due to missing dependency
when generating a javascript client, the generated .babelrc file is not valid json `globals` should be in quotes:

```
{
  "plugins": [
    ["babel-plugin-transform-builtin-extend", {
      globals: ["Error", "Array"]
    }]
  ],
  "presets": ["env", "stage-0"]
}
```
also, when I run `npm test` the test do not successfully compile with babel due to a missing dev dependency: `babel-plugin-transform-builtin-extend` which is referenced in the .babelrc file.

xyclos:
add quotes in .babelrc.mustache, so it generates valid json
add missing dev dependency in package.mustache, so babel can compile tests

fixes #638


johnteske:
I'm having this issue as well and have verified this—are you able to merge in the latest upstream master?

HugoMario:
thanks a lot @xyclos !!

PR #641, opened 2020-04-05
Go client codegen

HugoMario:
added client generator for golang


toonsevrin:
Waaw great job! Finally go support!!

How do I start using this?
```shell script
docker run --rm -v ${PWD}:/local swaggerapi/swagger-codegen-cli-v3 generate -i <file> -l go -o /local/openapi
```
Says that go is missing.

toonsevrin:
Hi @HugoMario, we are using a microsoft api that expects an EDM date format, but the generator passes it's time.time with the default formatting. What to do here?

HugoMario:
Hello @toonsevrin i think we could add an option to generator to work with the expected format. can you please file an issue for that and ping me there?

I'm not so familiar with go, so please, let me know what would be the expected output.

PR #642, opened 2020-04-09
Dot net 3 support

HugoMario:
Updated aspnetcore generator to work with aspnetcore v3


phil-warner:
Now that this is merged, how do we generate a .NET Core 3 client? The documentation doesn't mention it.

Issue #645, opened 2020-04-14
[javascript] model member keys are not quoted


johnteske:
Model members are not quoted, causing issues with kebab-case keys. Example:

```
export default class MyModel {
    ...
    /**
    * @member {Number} kebab-key
    */
   kebab-key = undefined;
}
```

PR #646, opened 2020-04-14
[javascript] quote model member keys

johnteske:
Addresses #645.


HugoMario:
nice catch, thanks a lot!!

PR #647, opened 2020-04-14
Fix java compile error due to ambiguous Object reference

davidgiga1993:
See swagger-api/swagger-codegen#10175

### DIFF
--- a/src/main/resources/handlebars/Java/libraries/okhttp-gson/JSON.mustache
+++ b/src/main/resources/handlebars/Java/libraries/okhttp-gson/JSON.mustache
@@ -31,6 +31,7 @@ import org.threeten.bp.format.DateTimeFormatter;
 
 import java.io.IOException;
 import java.io.StringReader;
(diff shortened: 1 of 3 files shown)


HugoMario:
hello @davidgiga1993, i checked the issue you reported, but i'm not sure this is the best approach for your issue fixing.

Other schemas, no named Object, will have this import.

There are a couple of propose in mind and would like to know what you think about it.

1. Rename a schema named Object to ModelObject. I mean to make the rename in output code not editing the document definition. This could be done including Object as a reserved word.
2. Update code and templates to call your schema named Object but it's full name, this is: package.name.Object

davidgiga1993:
Hi, 
yes I initally thought this was a general issue unitl I noticed it's caused by `Object` beeing in the schema itself. So this solution is not a good fit at all in that case.

The first solution seems to be the best as it could also easily used for other reserved keywords of the target languages. Also it reduces confusion as you directly see which type is used.

Issue #654, opened 2020-04-20
Fix for allOf discriminator issue breaks oneOf discriminator generation


trohrberg:
A couple of days ago, I realized that my project using the Swagger Codegen does not work anymore when using the current release version of Swagger Codegen. The problem is that, for a oneOf discriminator mapping, the corresponding Type-Info-Annotation is now missing in the POJO model classes:

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "typ", visible = true )
@JsonSubTypes({
}}

After investigating the issue further, I realized that the changes for PR #620 introduced a change which breaks that part of the code generation. Thus, I am able to provide a PR to fix that problem shortly.

HugoMario:
fixed wth #773

PR #655, opened 2020-04-20
Fix issue #654: Set discriminator for oneOf mapping correctly.

trohrberg:
Unfortunately, the changes of PR #620 introduced a problem for POJO model classes that contain oneOf discriminator mappings. The corresponding Type-Info-Annotation is not generated anymore:

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "typ", visible = true )
@JsonSubTypes({
}}

This PR fixes the implementation of DefaultCodeGeneratorConfig to make the generator generate the above part again.


dfrommi:
Thank you very much for the fix. I just came across the same issue and hope that your PR will soon be merged.

Until then, there's a hack to fix the issue. 
The discriminator is available if the type has an `allOf` definition. So you can extract properties to an artificial parent type.
The type with the discriminator will then have an `allOf` and an `anyOf`. Weird but working until your fix is part of a release.

HugoMario:
thanks a lot @trohrberg !

PR #658, opened 2020-04-23
Add the possibility for the RestTemplate library to enable collecting generic URI metrics

mariusneo:
Add the possibility for the RestTemplate library to enable collecting  metrics with URI template instead of the concrete URI.

Currently when doing a call to the resource path:

/shipments/{shipmentId}

with the shipmentId 123, the spring boot metric "http.client.requests" will contain the tag "uri" with the value "shipments/123".

The current change makes the resttemplate calls with appropriate uri variables when needed allowing therefor having the template "uri" tag values (e.g. : "shipment/{shipmentId}" in the previously mentioned example) for the spring boot metric "http.client.requests".


mariusneo:
should I add corresponding changes for the handlebars templates?

PR #664, opened 2020-04-27
Fixes #663. Remove 'Configuration' constructor sideeffect

### LINKED ISSUE #663: C# generated code default constructor of Client.Configuration throws TypeInitializationException
Generated C# code from OpenAPI 3.0 spec. Tried to create Configuration with the API endpoint and credentials to access the server. The generated client C# SDK throws an exception on the Client.Configuration constructor.

The code that reproduces the issue:

`var config = new Configuration();`

Result

` Message: 
    System.TypeInitializationException : The type initializer for 'IO.Swagger.Client.Configuration' threw an exception.
      ----> System.UriFormatException : Invalid URI: The format of the URI could not be determined.
  Stack Trace: 
    Configuration.ctor() line 91
    Tests.Test1() line 16
    --UriFormatException
    Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind)
    Uri.ctor(String uriString)
    RestClient.ctor(String baseUrl)
    ApiClient.ctor(String basePath) line 80
    Configuration.CreateApiClient() line 411
    Configuration.get_ApiClient() line 219
    Configuration.set_Timeout(Int32 value) line 252
    Configuration.ctor() line 122
    GlobalConfiguration.ctor()
    Configuration.cctor() line 108`

Root cause:

The reason for the above failure is the `Configuration` constructor has side-effect. It sets default value for the `Timeout` property which triggers creation of `ApiClient` with default value for `BasePath`. The default base path in the generated code is `"/"`. The latter is passed to the constructor of `RestClient` which expects a valid absolute Uri. 
The exception is thrown by `RestClient(string)` constructor which validates for a valid absolute Uri.

Suggested fix:
My suggestion is to remove the side-effect of  the default `Configuration` constructor.

dmilov:
Fixes #663 
Configuration type default constructor shouldn't have side effects. The issue with this side effect is indirect creation of a third party class `RestClient` with invalid argument for its `baseUrl` parameter.

The suggested fix removes create of `ApiClient` by the `Timeout` public property. `Timeout` property uses private field to hold the `Configuration` timeout value. ApiClient will be created on demand with the configuration instance which carries the value of the timeout. If `ApiClient` is created the `Timeout` property will update the underlying `ApiClient` instance.

Testing done:

1. Generated C# Client SDK from with the suggested fix.
2. Created `Configuration` instance with default constructor in the tests.

Result:
No exception.

3. Set timeout to configuration and check ApiClient was updated


dmilov:
@mandrean @HugoMario would you review the PR?

HugoMario:
thanks a lot @dmilov

PR #670, opened 2020-05-09
[typescript-fetch]: add supporting file for JSON decoding

johnteske:
This adds support for generating JSON decoders to ensure response types can be validated.


HugoMario:
Hello @johnteske, what is the purpose of adding an empty template?

johnteske:
> Hello @johnteske, what is the purpose of adding an empty template?

Apologies for the vague PR, I have updated the template and can give some more detail.

This should add basic support for generating JSON decoders using `ts.data.json`. I'm not able to address all the issues right now in how the types and decoders are generated but I hope this lays the foundation for future work, or that it at least gives an example of a JSON decoding solution.

Issue #673, opened 2020-05-11
embeddedTemplateDir empty when using custom templates


mmedjahed:
when I try to define custom templates, the generator does not get the embedded templates
When analyzing the generator code, I noticed that the embeddedTemplateDir variable is not initialized in this case 

[INFO] Skipped overwriting gradle.properties as the file already exists in /home/mmedjahed/basic-rest-client/target/generated-sources//gradle.properties
[INFO] Skipped overwriting AndroidManifest.xml as the file already exists in /home/mmedjahed/basic-rest-client/target/generated-sources/src/main/AndroidManifest.xml
[ERROR] 
java.lang.RuntimeException: Could not generate model 'AuthRequest'
	at io.swagger.codegen.v3.DefaultGenerator.generateModels(DefaultGenerator.java:451)
	at io.swagger.codegen.v3.DefaultGenerator.generate(DefaultGenerator.java:779)
	at io.swagger.codegen.v3.maven.plugin.CodeGenMojo.execute(CodeGenMojo.java:547)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:134)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:207)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:153)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:145)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:116)
	at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:80)
	at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build(SingleThreadedBuilder.java:51)
	at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:128)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:307)
	at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:193)
	at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:106)
	at org.apache.maven.cli.MavenCli.execute(MavenCli.java:863)
	at org.apache.maven.cli.MavenCli.doMain(MavenCli.java:288)
[...]

PR #674, opened 2020-05-11
#10107 (swagger-codegen) Combine custom templates and embedded templates

mmedjahed:
embeddedTemplateDir empty when using custom templates #673


mmedjahed:
#673 embeddedTemplateDir empty when using custom templates
Cannot use custom templates with swagger-codegen-v3 (https://github.com/swagger-api/swagger-codegen/issues/10107)

PR #703, opened 2020-06-19
added logback xml config file

HugoMario:


jyrimatti:
What was the reasoning behind this? I don't think libraries should include a logback.xml in the classpath...?

PR #713, opened 2020-06-30
added option to throw error instead return null for unknown enum values

HugoMario:
applied changes from [swagger-codegen#10356](https://github.com/swagger-api/swagger-codegen/pull/10356)


HugoMario:
please build

CDPrete:
Hi @HugoMario
I've tried specifying the `errorOnUnknownEnum` as `configOption` as well as `configuration` option in the `swagger-codegen-maven-plugin` but somehow it's not picked up (enums are still generated with a `fromValue` that returns `null` on a missing match).

How should this be applied? I'm targeting version 2.4.19 of the plugin.

Issue #719, opened 2020-07-03
[kotlin client] generates a deprecated library & old version of kotlin


dllewellyn:
``` compile "org.jetbrains.kotlin:kotlin-stdlib-jre8:$kotlin_version" ```
```ext.kotlin_version = '1.3.50'```

![Screenshot 2020-07-03 at 15 13 33](https://user-images.githubusercontent.com/12224340/86477166-c7bf1700-bd3f-11ea-9c4b-16a09991d33b.png)

Steps:

* Generate a kotlin client 
* Observe deprecated

PR #720, opened 2020-07-03
Updating kotlin version and updating dependency

dllewellyn:
Bumping kotlin version number and updating dependency

#719


HugoMario:
thanks @dllewellyn

PR #723, opened 2020-07-09
Fixes #722

### LINKED ISSUE #722: [Java] Wrong form parameters
Forwarded from the relevant [swagger-codegen issue](https://github.com/swagger-api/swagger-codegen/issues/10371).

##### Description

Something seems to have broken since version `3.0.20` with regards to form parameters because they are no longer generated as arguments to the api methods. Here is an example using `petstore.yaml`, comparing the java client SDK form method `updatePetWithForm` generated with version 3.0.20 to a java client SDK form method `updatePetWithForm` generated with the [latest commit](https://github.com/swagger-api/swagger-codegen/commit/3c955d7ce9055eb17084bbf0bec97d0c83c119af)

```diff
     /**
      * Build call for updatePetWithForm
      * @param petId ID of pet that needs to be updated (required)
-     * @param name  (optional)
-     * @param status  (optional)
      * @param progressListener Progress listener
      * @param progressRequestListener Progress request listener
      * @return Call to execute
      * @throws ApiException If fail to serialize the request body object
      */
-    public com.squareup.okhttp.Call updatePetWithFormCall(Long petId, String name, String status, final ProgressResponseBody.ProgressListener progressListener, final ProgressRequestBody.ProgressRequestListener progressRequestListener) throws ApiException {
+    public com.squareup.okhttp.Call updatePetWithFormCall(Long petId, final ProgressResponseBody.ProgressListener progressListener, final ProgressRequestBody.ProgressRequestListener progressRequestListener) throws ApiException {
```

As you can see, the form variables `name` and `status` have been ignored.

The corresponding method in [`petstore.yaml`](https://github.com/swagger-api/swagger-codegen/blob/3.0.0/modules/swagger-codegen/src/test/resources/3_0_0/petstore.yaml) file looks like this:
https://github.com/swagger-api/swagger-codegen/blob/3c955d7ce9055eb17084bbf0bec97d0c83c119af/modules/swagger-codegen/src/test/resources/3_0_0/petstore.yaml#L181-L208

##### Swagger-codegen version

I am using the latest version of swagger-codegen [as of today](https://github.com/swagger-api/swagger-codegen/commit/3c955d7ce9055eb17084bbf0bec97d0c83c119af)

For the comparison `diff` above, I compared it to the homebrew `swagger-codegen` command line tool, which is version 3.0.20.

##### Steps to reproduce

It is quick and easy to reproduce by comparing to version 3.0.19 instead of comparing to version 3.0.20 (will have the same result)

After clone and building the 3.0.0 branch repo,

```
java -jar modules/swagger-codegen-cli/target/swagger-codegen-cli.jar generate -i modules/swagger-codegen/src/test/resources/3_0_0/petstore.yaml -o samples/client/petstore/java/okhttp-gson  -l java
git diff samples/client/petstore/java/okhttp-gson/src/main/java/io/swagger/client/api/PetApi.java
```

This works because the latest version of `PetApi.java` in the repository [is from version 3.0.19](https://github.com/swagger-api/swagger-codegen/blob/3.0.0/samples/client/petstore/java/okhttp-gson/.swagger-codegen/VERSION)

##### Related issues/PRs

This PR #671  seems to fix this issue for typescript angular. A similar fix may be needed for java (and all it's respective libraries).

Edit: Upon further investigations, I believe the error actually lies somewhere in the transition from v1.0.20 to v1.0.21 in [swagger-codegen-generators](https://github.com/swagger-api/swagger-codegen-generators). I checked out the `3.0.20` version of this repo, and modified the `pom.xml` file here:

https://github.com/swagger-api/swagger-codegen/blob/491044d31c72b05ea24e15c843b1cda81f42e5ab/pom.xml#L1164

Changing from version `1.0.20` to `1.0.21-SNAPSHOT` and this results in the form parameters error described above.

UnHumbleBen:
This [commit](https://github.com/swagger-api/swagger-codegen-generators/commit/3ac195e7836baf6b081a17ecdb7c24f6e9c3f3fb#diff-371437d4e9e09ea18f2ff29dbacd0e84L4318-R4319) seems to forgotten to add formParams for `addCodegenContentParameters` function


HugoMario:
hey @UnHumbleBen thanks for your PR, this is something interesting to discuss. Maybe the method needs more refactor. I think form parameters and and body parameters are excluding each others according to :
https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java#L2149-L2191

So based on that:
`addParameters(content, codegenOperation.bodyParams);`
and
`addParameters(content, codegenOperation.formParams);`

should be called sequentially?

UnHumbleBen:
Can you explain more on what the issue is? My understanding is that if there are no bodyParams, then `addParameters(content, codegenOperation.bodyParams);` does nothing. If there are no formParams, `addParameters(content, codegenOperation.formParams);` does nothing.

HugoMario:
yea, you're right, that's what we expect.

PR #732, opened 2020-07-20
Added jackson imports to model for jaxrs-spec as enums use the annotations

mcassidygamma:
I encountered an issue when using the jaxrs-spec generator. For a pojo which used an enum, the enum was trying to use @JsonCreator and @JsonValue which were never being imported, thus compilation would fail.

I've added the imports and added a test also


mfatihercik (approved):
I encounter the same issue and it fixed the issue.

Issue #731, opened 2020-07-20
NPE in JavaClientCodegen when having "#/components/" prefixes


jerdno:
In `JavaClientCodegen.modelInheritanceSupportInGson()` there is iteration over `byParent.entrySet()`. Inside the iteration keys are being modified by `parentModel.discriminator.getMapping().replaceAll((key, value) -> OpenAPIUtil.getSimpleRef(value))` which is against contract and are producing inconsistent results. 

For example `byParent.get(parentModel)` can return null if parent of current parentModel was already modified in previous iteration. After that there is NPE on `for (CodegenModel model : childrenModels)`, but root cause is modification of keys.

PR #733, opened 2020-07-24
NPE in JavaClientCodegen when having "#/components/" prefixes #731

rbrisuda:
Schemas are loaded in alphabetic order. If this order is wrong, logic in: modelInheritanceSupportInGson method can fail on NPE with reason that keys (their hascodes) in map are modified during for cycle iteration.


HugoMario:
thanks a lot @rbrisuda !!

Issue #741, opened 2020-08-12
generate spring with oneOf & serializableModel


Fyro-Ing:
Hello

I use code gen generators 1.016 for spring with serializableModel = true (false is ok)

 Upgrading to last (or 1.0.17) failed with additional comma 

openapi.yaml
```
        homePhone:
          oneOf:
            - $ref: '#/components/schemas/label'
            - $ref: '#/components/schemas/encrypted-label'
        mobilePhone:
          oneOf:
            - $ref: '#/components/schemas/label'
            - $ref: '#/components/schemas/encrypted-label'
```
```
public class EncryptedLabel  implements Serializable , OneOfpersonHomePhone, , OneOfpersonMobilePhone, , OneOfpersonProfessionalPhone  {
```

regression introduced with #590  ?

calbach:
I think https://github.com/swagger-api/swagger-codegen-generators/pull/743 should fix.

PR #743, opened 2020-08-25
Fix Java codegen for multi-interface Serializable models

calbach:
Fixes https://github.com/swagger-api/swagger-codegen-generators/issues/741

Previous codegen resulted in:

```
public class MyModel implements Serializable, OneOfMyInterface1, , OneOfMyInterface2 {
  ...
```


Fyro-Ing:
any update to valid this fix ?

Issue #686, opened 2020-05-20
JsonTypeInfo not generated for discriminator in 3.0.20


pschichtel:
I just tried to upgrade a project from 3.0.16 to 3.0.20, but after that for our result types no @JsonTypeInfo and @JsonSubType annotations are being generated, forcing me to roll back to the old version, which still works fine.

Reduced example of the API definition:

```yaml
openapi: 3.0.0
servers:
info:
security:
components:
  schemas:
    ProvisionCallResponse:
      type: object
      required:
        - type
      properties:
        type:
          type: string
          enum:
            - Success
            - Error
          description: the descriminator
      discriminator:
        propertyName: type

    Success:
      allOf:
        - $ref: '#/components/schemas/ProvisionCallResponse'
        - type: object
          required:
            - target
          properties:
            target:
              type: string
              description: the target phone number to be called
            dialogId:
              type: string
              format: uuid
              description: the dialog ID for accessing dialog history data

    Error:
      allOf:
        - $ref: '#/components/schemas/ProvisionCallResponse'
        - type: object
          required:
            - message
          properties:
            message:
              type: string
              description: the message describing the error
```

3.0.16 generates the following ProvisionCallResponse class:

```java
/**
 * ProvisionCallResponseModel
 */
@Validated
@javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.SpringCodegen", date = "2020-05-19T16:39:44.201307500+02:00[Europe/Berlin]")
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type", visible = true )
@JsonSubTypes({
  @JsonSubTypes.Type(value = ErrorModel.class, name = "Error"),
  @JsonSubTypes.Type(value = SuccessModel.class, name = "Success"),
})
public class ProvisionCallResponseModel   {
  // stuff
}
```

3.0.20 generates the following ProvisionCallResponse class:

```java
/**
 * ProvisionCallResponseModel
 */
@Validated
@javax.annotation.Generated(value = "io.swagger.codegen.v3.generators.java.SpringCodegen", date = "2020-05-19T16:39:44.201307500+02:00[Europe/Berlin]")
public class ProvisionCallResponseModel   {
  // stuff
}
```

[...]

PR #773, opened 2020-10-09
fix: bring back discriminator initialization to where it was before

kaibra:
PR: https://github.com/swagger-api/swagger-codegen-generators/pull/620 , which was introduced in `1.0.19`, seems to be causing multiple issues (eg https://github.com/swagger-api/swagger-codegen-generators/issues/654).

This PR aims at fixing two issues:
fixes https://github.com/swagger-api/swagger-codegen-generators/issues/686
fixes https://github.com/swagger-api/swagger-codegen/issues/10198

Both issues are about missing polymorphic jackson annotations (`JsonSubTypes`, `JsonTypeInfo`).

The issue was caused by moving the initialization of the `discriminator` into the `ComposedSchema` block (https://github.com/swagger-api/swagger-codegen-generators/pull/620/commits/5d2238265a02d0ee3f19dc8dc9b4ed1e5d970b8c#diff-371437d4e9e09ea18f2ff29dbacd0e84L1279). 

In the examples mentioned in the issues, the model is not a `ComposedSchema`.
But the `discriminator` needs to be initialized for the annotations to be generated, but it is only for `ComposedSchema`s now.

This PR moves the initialization back to where it was before.
I am not quite sure if this has other side-effects, I did not see. It looks like there are a bunch of cases untested?!
Still I hope this might speed up things in case it gets approved.

Cheers,

Kai

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
@@ -1333,6 +1333,8 @@ public abstract class DefaultCodegenConfig implements CodegenConfig {
         }
         codegenModel.getVendorExtensions().put(CodegenConstants.IS_ALIAS_EXT_NAME, typeAliases.containsKey(name));
 
+        codegenModel.discriminator = schema.getDiscriminator();
+
         if (schema.getXml() != null) {
             codegenModel.xmlPrefix = schema.getXml().getPrefix();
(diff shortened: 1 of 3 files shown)


asauray:
This 👍 

I hope somebody can review it soon

asauray:
@HugoMario Could you take a look ? It seems you have worked on a related issue earlier last summer.

HugoMario:
thanks a lot @kaibra for this PR!!!

PR #838, opened 2021-01-08
Make sure we don't get two entries in codegenContents with the same schema

### LINKED ISSUE #837: Duplicate codegenContents when requestBody has multiple content types and one of them is a form
Given a resource with multiple content types, the generated code will not compile if one or more of these content types is either `application/x-www-form-urlencoded` or `multipart/form-data`.

In the case where the requestBody is identical, the following fix seems to work. When the request bodies are different, you'll hit #441.

alf-mindshift:
Given the same schema, the templates will generate identical operations. In some languages this is fine, but in for instance Java this will cause the compiler to barf. 

I don't have time to make a failing tests, but that should be trivial. This pull fixes #837.


alf-mindshift:
This fix is not good enough since the resulting operation will look different depending on the order of the content types.

PR #839, opened 2021-01-08
[Micronaut] remove 'defaultValue' in @Parameter annotations in Micronaut templates

pbthorste:
If the swagger file has a default value set on a parameter, a 'defaultValue' will be added to the @Parameter annotation. 
The problem is that this annotation uses `io.swagger.v3.oas.annotations.Parameter` does not support this field.


pbthorste:
Hi @HugoMario  - any chance of getting this PR merged ?

PR #841, opened 2021-01-11
[Python] Adding type-hinting for API responses

Dringho:
Tested in Pycharm and vscode

image

### DIFF
--- a/src/main/resources/handlebars/python/__init__api.mustache
+++ b/src/main/resources/handlebars/python/__init__api.mustache
@@ -5,3 +5,5 @@ from __future__ import absolute_import
 # import apis into api package
 {{#apiInfo}}{{#apis}}from {{importPath}} import {{classname}}
 {{/apis}}{{/apiInfo}}
+{{#models}}{{#model}}from {{modelPackage}}.{{classFilename}} import {{classname}}
+{{/model}}{{/models}}
--- a/src/main/resources/handlebars/python/__init__api.mustache
+++ b/src/main/resources/handlebars/python/__init__api.mustache
@@ -2,8 +2,10 @@ from __future__ import absolute_import
 
 # flake8: noqa
 
+# Import models for type-hinting
+{{#models}}{{#model}}from {{modelPackage}}.{{classFilename}} import {{classname}}
+{{/model}}{{/models}}
+
 # import apis into api package
 {{#apiInfo}}{{#apis}}from {{importPath}} import {{classname}}
 {{/apis}}{{/apiInfo}}
-{{#models}}{{#model}}from {{modelPackage}}.{{classFilename}} import {{classname}}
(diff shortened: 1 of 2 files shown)


HugoMario:
Hey @Dringho , why you closed this? Need any help?

Dringho:
@HugoMario I was trying to improve the solution but I couldn't without delving into the Java, which was a mess to understand.

My current problem it's that on the generated inline_response_200 response, if you have a nested structure, it references to answers like inline_response_200_message, but there was no way to add the import for it, so the type hinting would be enabled at the nested level.

I was working on the `model.mustache` file adding imports like this
```
# Importing related models
{{#models}}{{#model}}{{#vars}}{{#complexType}}from .I_CANT_GET_THIS_RIGHT import {{complexType}}
{{/complexType}}{{/vars}}{{/model}}{{/models}}
```

To fix this, it would be needed to add the "importPath" or "filename" key to the nested Array/Object schema objects, but that fix would require a better knowledge of the Java code that I don't currently have.

MalteEbner:
Would it be possible to already have a basic typehinting functionality for types like 'float', 'dict'... that just gives 'object' as a type hint instead of importing the more precise type?

Dringho:
> Would it be possible to already have a basic typehinting functionality for types like 'float', 'dict'... that just gives 'object' as a type hint instead of importing the more precise type?

I'm not sure I understand your idea. Why would you type hint something broader than the actual return? With this PR the basic typehints work as well as the complex ones (except for the nested complex types)

MalteEbner:
> > Would it be possible to already have a basic typehinting functionality for types like 'float', 'dict'... that just gives 'object' as a type hint instead of importing the more precise type?
> 
> I'm not sure I understand your idea. Why would you type hint something broader than the actual return? With this PR the basic typehints work as well as the complex ones (except for the nested complex types)

If I understand it correctly we have 4 possibilities how many typehint functionality to have (listed in increasing value):

1. None (like currently on master)
2. Basic (like I proposed in my last comment)
3. Basic + complex (like in your branch)
4. Basic + complex + nested-complex (which would be best)

I assume that the branch is not merged because getting 4) to work is quite complex.
As I currently only have 1), I would prefer to have 2) or 3) merged, than trying to get 4) working.

MalteEbner:
@Dringho Do you know why this PR is not approved? Does it need to be rebased on master or are there additional problems?

Dringho:
@MalteEbner I do not know why this Pull request has not been reviewed/approved. I'm not sure about the development cylce of this project, I'm just an outside contributor.
@HugoMario would you care to provide some feedback here?

dangell7:
BUMP

dangell7:
Can someone rebase and merge this? Who wants to make some $$?

PR #821, opened 2020-12-10
[PHP] FIX ObjectSerializer.mustache

r-simlinger:
In https://github.com/swagger-api/swagger-codegen-generators/commit/826dc2aa2368c6c0c3bbe380d2406eed97154ee0 the inner self call was not adjusted.

PR #866, opened 2021-02-19
changed order of arguments

a4blue:
#865


r-simlinger:
It's not the wrong order. The issue is that the method has only 2 arguments, but the call set there 3. So you need to remove one: https://github.com/swagger-api/swagger-codegen-generators/pull/821

PR #871, opened 2021-03-03
jackson update to 2.11.4

frantuma:


dhpalan:
@frantuma 
Is there a reason why the `jackson-version` here is not updated?

https://github.com/swagger-api/swagger-codegen-generators/blob/7ad4d49f6db062e2de312a0832afc548b4b0bb63/src/main/resources/handlebars/Java/libraries/resttemplate/pom.mustache#L284

PR #875, opened 2021-03-09
Enable StringHelpers Handlebars in Kotlin client codegen

poculeka:
Registering `StringHelpers` for Kotlin client codegen will help users generating code with custom Handlebars templates. For example, Retrofit HTTP client is requiring endpoint paths to skip leading slash to become relative. If defined server URL is `http://server.com/gateway/` and endpoint path is `/api/endpoint/`, then full path becomes `https://server.com/api/endpoint` because of leading slash in the latter. This is a great use case to apply Handlebars helper `{{substring path 1 }}` as URL `https://server.com/gateway/` and path `api/endpoint` would join relatively and produce expected `https://server.com/gateway/api/endoint`.


marcin-jelenski:
@HugoMario any chances to merge?

marcin-jelenski:
Thanks @HugoMario ✌️

PR #877, opened 2021-03-12
[Typescript] Set parent for additional properties to Record<string, Child>

### LINKED ISSUE #691: Typescript instantiationType for maps
Hi!

I have an issue where an CodegenModel for a map-like object is incorrectly generated. I am not sure which way to go from here. Is it the generated OASv3 model or the code generator for typescript. I would appreciate any guidance from this community.

I generated a specification for a spring boot project of mine with springdoc. There I use spring data rest which generates APIs with Link/Links. There is a customization in [springdoc](https://github.com/springdoc/springdoc-openapi/blob/v1.3.9/springdoc-openapi-data-rest/src/main/java/org/springdoc/data/rest/SpringDocDataRestConfiguration.java#L103) for these:
```java
		ResolvedSchema resolvedLinkSchema = ModelConverters.getInstance()
				.resolveAsResolvedSchema(new AnnotatedType(Link.class));

		return openApi -> openApi
				.schema("Link", resolvedLinkSchema.schema)
				.schema("Links", new MapSchema()
						.additionalProperties(new StringSchema())
						.additionalProperties(new ObjectSchema().$ref("#/components/schemas/Link")));
```

This is the excerpt for just the links.
```
class MapSchema {
        class Schema {
            type: object
            format: null
            $ref: null
            description: null
            title: null
            multipleOf: null
            maximum: null
            exclusiveMaximum: null
            minimum: null
            exclusiveMinimum: null
            maxLength: null
            minLength: null
            pattern: null
            maxItems: null
            minItems: null
            uniqueItems: null
            maxProperties: null
            minProperties: null
            required: null
            not: null
            properties: null
            additionalProperties: class Schema {
                type: null
                format: null
                $ref: #/components/schemas/Link
                description: null
                title: null
                multipleOf: null
                maximum: null
                exclusiveMaximum: null
                minimum: null
                exclusiveMinimum: null
                maxLength: null
                minLength: null
                pattern: null
                maxItems: null
                minItems: null
                uniqueItems: null
                maxProperties: null
                minProperties: null
                required: null
                not: null
                properties: null
                additionalProperties: null
                nullable: null
                readOnly: null
                writeOnly: null
                example: null
                externalDocs: null
                deprecated: null
                discriminator: null
                xml: null
            }
            nullable: null
            readOnly: null
            writeOnly: null
            example: null
            externalDocs: null
            deprecated: null
            discriminator: null
            xml: null
        }
    }
```

When generating the typescript-fetch template I get to the method `io.swagger.codegen.v3.generators.DefaultCodegenConfig#toInstantiationType`. This method gets the `map` key from `instantiationTypes` as seen here https://github.com/swagger-api/swagger-codegen-generators/blob/76ea8a89df73928b10e0c84f1c8356739e34bf3d/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java#L1021

This is `null` because the `map` key is never set seen here https://github.com/swagger-api/swagger-codegen-generators/blob/2a561d8587e5bea502941e6c5377ce89eff6f13d/src/main/java/io/swagger/codegen/v3/generators/typescript/AbstractTypeScriptClientCodegen.java#L86

So CodegenModel.parent becomes `null<String, Link>` which causes the generated model to be like this:
```typescript
export interface Links extends null<String, Link> {
    [key: string]: Link;
}
```

I think the proper typescript way would just be:
```typescript
export interface Links {
    [key: string]: Link;
}
```

or maybe alternatively, keeping extends:
```typescript
export interface Links extends Map<String, Link> {
    [key: string]: Link;
}
```

There are some issues regarding the springdoc generated OASv3 spec like https://github.com/springdoc/springdoc-openapi/issues/524.

Lukas-Kullmann:
This fixes #691

The gist of the issue is that currently, the typescript generators generate additional properties to have a parent of `null<String, Child>`. This is wrong in two dimensions. 

1. The parent type should not be `null`. I chose [`Record`](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) in this PR as a default.
2. The parameter of the key is in upper case. This is wrong in Typescript and has to be lower case. This actually comes hard-coded from the DefaultCodegenConfig class. I guess this is a bug here, but I did not dare touching it since it affects all languages.

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/typescript/AbstractTypeScriptClientCodegen.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/typescript/AbstractTypeScriptClientCodegen.java
@@ -76,7 +76,8 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegenConf
                 "any",
                 "File",
                 "Error",
-                "Map"
+                "Map",
+                "Record"
                 ));
 
         languageGenericTypes = new HashSet<String>(Arrays.asList(
@@ -84,6 +85,7 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegenConf
         ));
 
         instantiationTypes.put("array", "Array");
+        instantiationTypes.put("map", "Record");
 
         typeMapping = new HashMap<String, String>();
         typeMapping.put("Array", "Array");
@@ -471,5 +473,16 @@ public abstract class AbstractTypeScriptClientCodegen extends DefaultCodegenConf
     @Override
     public ISchemaHandler getSchemaHandler() {
         return new TypeScriptSchemaHandler(this);
+    }
+
+    @Override
+    public String toInstantiationType(Schema property) {
+        String instantiationType = super.toInstantiationType(property);
+
+        if (instantiationType != null && instantiationType.contains("<String, ")) {
+            return instantiationType.replaceFirst("<String, ", "<string, ");
+        }
+
+        return instantiationType;
     }
 }


filiplindqvist-netlight:
Seems resonable to me. At the very least `map=Record` should be default. Workaround while waiting for this PR is adding `--instantiation-types map=Record` to the CLI. However, this still generates `Record<String, `instead of `Record<string, `.

Paku580:
Are there any updates on this? This fix would help me very much too

PR #896, opened 2021-04-30
Fix for stripping prefix on single enums

barryadk:
This PR resolves a bug when using a single enum is used that a "common prefix" is found resulting in the enum's name being cropped: https://github.com/swagger-api/swagger-codegen/issues/7725

The above issues was fixed in Codegen OAS 2, this PR applies the same fix https://github.com/swagger-api/swagger-codegen/pull/7726, plus the addition of unit tests.


HugoMario:
thanks @barryadk !!

PR #899, opened 2021-05-01
[typescript-fetch] use isomorphic-fetch

mildsunrise:
For `typescript-fetch`, the generated client uses the [`portable-fetch`](https://www.npmjs.com/package/portable-fetch) package. It's a fork of the very popular [`isomorphic-fetch`](https://www.npmjs.com/package/isomorphic-fetch) package. The fork has been unmaintained for 4 years and provides no documentation on its motives.

This PR replaces the fork with the original (more up to date) package. I think it's not good for codegen to rely on random forks for such simple tasks (a polyfill for `fetch()`). If users have specific needs regarding the `fetch()` implementation, the generated code allows them to pass it on the constructor.

### DIFF
--- a/src/main/resources/handlebars/typescript-fetch/api.mustache
+++ b/src/main/resources/handlebars/typescript-fetch/api.mustache
@@ -3,7 +3,7 @@
 {{>licenseInfo}}
 
 import * as url from "url";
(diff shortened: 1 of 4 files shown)


HugoMario:
Awesome, thanks a lot!

PR #900, opened 2021-05-03
[typescript] fix type aliases generation

mildsunrise:
For schemas of primitive types:

~~~ yaml
components:
  schemas:
    Timestamp:
      description: UTC timestamp in milliseconds since epoch
      type: number
~~~

It was previously generating invalid code:

~~~ typescript
/**
 * UTC timestamp in milliseconds since epoch
 * @export
 */
export interface Timestamp {
}
~~~

Now it generates:

~~~ typescript
/**
 * UTC timestamp in milliseconds since epoch
 * @export
 */
export type Timestamp = number
~~~


mildsunrise:
(also applies to arrays)

HugoMario:
LGTM, thanks a lot @mildsunrise !!!

PR #910, opened 2021-05-28
909: Fixed bug where an enum value could end up being called "_" whic…

### LINKED ISSUE #909: [Java] enum values can be called _ which does not compile in Java 11
##### Description

If you declare a string enum, swagger-codegen tries to convert the value into an identifier. This may result in an identifier called just "_". This is a valid identifier in Java <= 8, but not in Java >= 9 (e.g. Java 11).

Java reports the error
```
as of release 9, '_' is a keyword, and may not be used as an identifier
```

Related issue shows a case where this happens, but it is a general problem with "_" identifiers and Java 11. https://github.com/swagger-api/swagger-codegen/issues/10493

##### Swagger-codegen version

Not a regression afaik.

##### Swagger declaration file content or url

```yaml
    FieldValueClause:
      required:
      - "operator"
      type: "object"
      properties:
        operator:
          type: "string"
          description: "The operator between the field and operand."
          enum:
          - "="
          - "~="
```

This generates code like

```java
  /**
   * The operator between the field and operand.
   */
  public enum OperatorEnum {
    EQUAL("="),
    _("~=");
  }
```

##### Command line used for generation

swagger-codegen generate \
    -i input.yaml \
    -l java \
    -o .

##### Steps to reproduce

Try to compile the resulting code with Java 11.

##### Related issues/PRs

This issue demonstrates the same problem. The identifier compiles in Java 8, but not Java 11. https://github.com/swagger-api/swagger-codegen/issues/10493

##### Suggest a fix/enhancement

A better way to make identifiers unique rather than prefix underscores, esp for Java >= 9.

bkelley17:
…h is invalid in Java >= 9.

Fixes #909

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaCodegen.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaCodegen.java
@@ -492,6 +492,15 @@ public abstract class AbstractJavaCodegen extends DefaultCodegenConfig {
         }
     }
 
+    protected String escapeUnderscore(String name) {
+        // Java 8 discourages naming things _, but Java 9 does not allow it.
+        if("_".equals(name)) {
+            return "_u";
+        } else {
+            return name;
+        }
+    }
+
     @Override
     public String escapeReservedWord(String name) {
         if(this.reservedWordsMappings().containsKey(name)) {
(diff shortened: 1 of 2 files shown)


bkelley17:
There was an existing check to prevent variables being called "_", but the same check was not applied to enums. This patch applies this to enums as well, so that the code will compile OK with Java 11.

HugoMario:
LGTM, thanks a lot @bkelley17 !!

PR #929, opened 2021-06-29
fixed NPE on spring generator when using discriminator mapping

HugoMario:


HugoMario:
fixed NPE on https://github.com/swagger-api/swagger-codegen/issues/11049

PR #930, opened 2021-07-06
Fixed #10129 for swagger-codegen where python was getting the type "O…

bkelley17:
…bject" when it should be "object". See https://github.com/swagger-api/swagger-codegen/issues/10129


HugoMario:
Thanks a lot @bkelley17 !!

Issue #939, opened 2021-08-10
[java] setter name wrong where child property clashes with parent property


bkelley17:
If a child property has the same name as the parent property, the getter for the child property will be called something like getChildTypeProperty, but the setter also gets called getChildTypeProperty (not setChildTypeProperty).

bkelley17:
I'll fix this at the same time as #936

PR #940, opened 2021-08-10
Fixes getter and setter name problems for subtypes

### LINKED ISSUE #936: [java] Property names that differ only by underscores have the same getter/setter name
A property called "_value" will have a getter of "getValue()" and a setter of "setValue()", as will a property called "value".
If a parent type has a property "value" and the child (allof the parent type) has a property "_value", then the resulting code won't compile.
The parent will have a getValue() method that returns the type of "value", and child will have a getValue() method that returns the type of "_value".
Possibly the getter/setter name generation should allow underscores in the method name for Java.

```
      "ParentType": {
        "description": "The parent type.",
          {
            "type": "object",
            "required": [
              "value"
            ],
            "additionalProperties": false,
            "properties": {
              "value": {
                "$ref": "#/components/schemas/ValueType"
              }
            }
          }
        ]
      },
      "ChildType": {
        "description": "The child type.",
        "allOf": [
          {
            "$ref": "#/components/schemas/ParentType"
          },
          {
            "type": "object",
            "required": [
              "_value"
            ],
            "additionalProperties": false,
            "properties": {
              "_value": {
                "$ref": "#/components/schemas/ValueUnderscoreType"
              }
            }
          }
        ]
      },

bkelley17:
This fixes #936 and #939 to do with how getters and setters are named when the subtype clashes with the parent type.

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaCodegen.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/java/AbstractJavaCodegen.java
@@ -33,7 +33,6 @@ import org.slf4j.LoggerFactory;
 
 import java.io.File;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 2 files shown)


HugoMario:
thanks @bkelley17 !!

PR #947, opened 2021-08-24
feat: update Java Micronaut generator to support Micronaut 3

### LINKED ISSUE #946: [JavaMicronaut] Micronaut 3 support
Micronaut 3 no [longer ships](https://docs.micronaut.io/3.0.0/guide/#_rxjava2) any reactive implementation as a default now.

Current Micronaut code generator generates code for RxJava 2, but it is [end-of-life](https://github.com/ReactiveX/RxJava/blob/b7f66c8aff0e6f1328c7c7e3d445279197d056b3/README.md) as of February 28, 2021. Micronaut itself now uses Project Reactor internally, and Micronaut developers recommend switching to it to avoid having multiple reactive implementations in classpath.

It would be good to have an option to choose Recative Streams implementation for generated code.

Paullo612:
- fix useBeanValidation option application
- add support for choosing Reactive Streams implementation for generated
  code through library option
- remove `@Controller` annotation from generated interfaces, as it is not
  `@Inherited` anyway
- use jakarta inject implementation instead of javax, as Micronaut now
  uses it by default
- generate stub controllers when skipSupportFiles is disabled to be able
  to run tests and start generated application
- fix bean validation annotations
- add cookies parameters support
- use `@Body` instead of `@QueryValue` for form parameters generation
- update pom.xml to use Micronaut 3

closes #946


src/main/java/io/swagger/codegen/v3/generators/java/MicronautCodegen.java
```diff
@@ -155,17 +169,18 @@ public void processOpts() {
             skipSupportFiles = Boolean.valueOf(additionalProperties.get(SKIP_SUPPORT_FILES).toString());
         }
 
-        if (useBeanValidation) {
-            writePropertyBack(USE_BEANVALIDATION, useBeanValidation);
-        }
+        writePropertyBack(USE_BEANVALIDATION, useBeanValidation);
```
Paullo612, line 172: Looks like handlebars only understands booleans, but not booleans represented as strings. So, we have to put boolean to `additionalProperties` instead of its string representation.

src/main/resources/handlebars/JavaMicronaut/api.mustache
```diff
@@ -10,11 +10,17 @@ package {{package}};
...
+{{#useRxJava2}}
 import io.reactivex.Single;
+{{/useRxJava2}}
+{{#useRxJava3}}
+import io.reactivex.rxjava3.core.Single;
+{{/useRxJava3}}
 import io.swagger.v3.oas.annotations.*;
 import io.swagger.v3.oas.annotations.responses.*;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
```
Paullo612, line 17: Looks like those was copy & pasted from another template. I do not see those to be used in generated code.

src/main/resources/handlebars/JavaMicronaut/beanValidationParamsInner.mustache
```diff
@@ -0,0 +1,5 @@
+{{^isPrimitiveType}}{{^isEnum}}{{!
+    Non-container
+}}{{^isContainer}}@Valid {{/isContainer}}{{!
+    Container
+}}{{#isContainer}}{{#items}}{{>beanValidationParamsInner}}{{/items}}{{/isContainer}}{{/isEnum}}{{/isPrimitiveType}}
```
Paullo612, line 5: Had to introduce recursive template :( But I do not see any other way to implement this. Looks like Micronaut throws `ConstraintViolationException` saying something about absent introspection if `List` of primitive types is marked as `@Valid`. This is there to handle `List<List<String>>` case.

Paullo612:
This should not brake anything, as I've left RxJava 2 as default Reactive Streams implementation. The only thing I'm not sure about is removal of `@Controller` from generated interfaces. Haven't checked if it was inherited to implementation classes with Micronaut 2. But Micronaut generator is marked as experimental anyway.

@HugoMario, can you please take a look at this.

HugoMario:
hey @Paullo612 
sirry for this long delay, I'm going to check this.

PR #954, opened 2021-09-08
feat: use default default rxjava3 adapter for retrofit2

chadoc:
When adding support of RxJava3 for Retrofit2 generator, adapter used was a third party library: https://github.com/swagger-api/swagger-codegen/issues/10312

An "official" adapter is provided by squareup itself which follows the versioning of retrofit2. It's probably a better choice to stick to default implementation.

Previously, RxJava3CallAdapterFactory.create() was returning an adapter which was creating synchronous Observables by default. Now, RxJava3CallAdapterFactory.create() returns an adapter which create async Observables by default, that's why I keep previous behaviour, however, maybe we should stick to default behaviour desired by squareup.


yokrysty:
proposal: use
`RxJava3CallAdapterFactory.createWithScheduler(Schedulers.io())`

yokrysty:
why is this still not fixed?

PR #997, opened 2021-12-19
added jackson annotations on generated interface for deserialization

HugoMario:


msmerc:
It's missing a `@Generated` annotation ;)

msmerc:
What happens if the api spec already has a field called `type`? The serialisation will overwrite this!

Issue #1001, opened 2021-12-28
Support for generating qualified model package/namespace


frode-carlsen:
The generator will currently place all generated model files in the same folder.  This doesn't work well with larger models that may support different versions of the same API (requiring all model files to have unique names). The translation of schema names to model names also does a significant amount of character substitution, changing case etc. of the schema names, which further separates the generated code from the assigned schema name.

For example the following
` /my/path:
    post:
      ....
      requestBody:
        ...
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/some.namespace.Model'`

will generate a `SomeNamespaceModel` class file in the `io.swagger.model` namespace instead of a `Model` class in `some.namespace`

Because there's a significant amount of conversion of model names by legacy (transliterating _. etc to title case. 
I propose to add a new flag to support generating fully-qualified names: `useFqn`

PR #1002, opened 2021-12-28
added useFqn parameter to generate fully qualified model class names (Java) (#1001)

frode-carlsen:


frode-carlsen:
This addresses #1001 by adding an additional property `useFqn` that will take the assigned schema name as the basis for the generated package namespace of the model files.  Any modelPackage already set (or default io.swagger.model) will be prepended to the package namespace.

This currently only supports the generated Java code, but could relatively easily be extended to the other language templates.

As an asside, a couple of refactorings ought be done to swagger-codegen to improve this.  If this seems a viable PR then I'd be happy to submit a PR for that too:
* DefaultGenerator#processModels only processes one model at a time atm.  The templates are complicated to support multiple models when only one is present at a time.  Propose simplifying this method signature to only take one definition at a time if possible 
* I've reused CodegenModel#xmlNamespace field for this case to better support generated import statement.  This should be generalized in swagger-codegen to a namespace field or similar

frode-carlsen:
@frantuma @HugoMario would it be possible to get feedback on this PR?  This would be very helpful for generating code from larger OpenAPI specs.

PR #1009, opened 2022-01-16
[DANGEROUS security problem] no escape for query parameters for Dart/Flutter language, so any query parameter with char like ?,&,= etc will have problems

### LINKED ISSUE #1008: [DANGEROUS security problem] no escape for query parameters for Dart/Flutter language, so any query parameter with char like `?`,`&`,`=` etc will have problems
https://github.com/swagger-api/swagger-codegen-generators/blob/698ed072a7392c26066b6d8d340d2086f7304670/src/main/resources/handlebars/dart/api_client.mustache#L108-L111

    var ps = queryParams.where((p) => p.value != null).map((p) => '${p.name}=${p.value}');
    String queryString = ps.isNotEmpty ?
                         '?' + ps.join('&') :
                         '';

This code has problems: For example, suppose the user inputs `a&b=c` for a parameter named `x`. Then it will be converted to `mywebsite.com/mypath?x=a&b=c`, and the server will receive a param named `x` with value `a` and *another param* named `b` with value `c`.

## Example security attack

Here is a concrete example of a very naive security attack. Suppose we are building a bank app and there is a money-transfer page. The page contains two input fields: A "amount" input (how much money to transfer) and a "target" input (the target bank account username). After user inputs the amount and the remark and click "ok" button, a HTTP request using swagger-codegen-generators will be sent. It is like the following: `bank.com/transfer?amount=123&target=bob`. In addition, that backend API is very powerful, and can accept `currency` as an argument. For example, `bank.com/transfer?amount=123&target=bob&currency=euro` uses euro instead of USD as a unit.

Suppose alice only wants to transfer $1 to bob. However, bob the attacker can tell alice, she should transfer $1 to an account named `bob_real_account_name&currency=euro` and let Alice believe it is just bob's account name (though a bit strange). Then what happens? If alice do this, she will transfer 1 euro (instead of 1 USD) to bob - she loses (extra) money!

Surely this is just a naive example, and with more thoughts we should be able to come up with more harmful examples. but this is enough to show that the bug can be very damaging!

Based on the "insights" page, @HugoMario @frantuma hi could you plz have a look? thanks!

fzyzcjy:
Closes #1008 

please see that issue for problem description

### DIFF
--- a/src/main/resources/handlebars/dart/api_client.mustache
+++ b/src/main/resources/handlebars/dart/api_client.mustache
@@ -105,7 +105,7 @@ class ApiClient {
 
     _updateParamsForAuth(authNames, queryParams, headerParams);
 
-    var ps = queryParams.where((p) => p.value != null).map((p) => '${p.name}=${p.value}');
+    var ps = queryParams.where((p) => p.value != null).map((p) => '${Uri.encodeComponent(p.name)}=${Uri.encodeComponent(p.value)}');
     String queryString = ps.isNotEmpty ?
                          '?' + ps.join('&') :
                          '';


frantuma:
Thanks!

fzyzcjy:
You are welcome!

PR #1016, opened 2022-02-23
Fix: replace wrong template variable readOnly with isReadOnly

fedepad:
In few templates, the template variable `readOnly` is used, which is not defined. 
Instead, `isReadOnly` is the defined template variable.
This PR fixes that by replacing occurrences of the undefined template variable `readOnly` with the defined `isReadOnly`.


fedepad:
Fix https://github.com/swagger-api/swagger-codegen-generators/issues/1017

PR #1022, opened 2022-03-10
[Java/Feign] Don't HTML escape Accept / Content-Type header

blutorange:
When e.g. the Accept-Header contains special characters such as `=`, mustache escapes them, which results in code like this to be generated:

```java
// MyApi.java
public interface MyApi extends ApiClient.Api {
  @RequestLine("GET /rest/protected/items?size={size}&page={page}")
  @Headers({
      "Content-Type: */*",
      "Accept: application/com.example+json; version&#x3D;1.0; charset&#x3D;utf-8,application/com.example+xml; version&#x3D;1.0; charset&#x3D;utf-8,application/json; charset&#x3D;utf-8,application/xml; charset&#x3D;utf-8",
  })
  LicenseListResource getLicenses(@QueryMap(encoded=true) Map<String, Object> queryParams);
}
```

Note the `Accept: application/com.example+json; version&#x3D;1.0; charset&#x3D;utf-8` with the escaped `=`, which does not work. This disables HTML escaping for the header values.

On a side note though, I had been wondering whether HTML escaping makes sense when generating Java files? It does in JavaDocs I suppose, but shouldn't it be disabled everywhere else?


HugoMario:
thanks @blutorange !

PR #1045, opened 2022-05-26
Fix missing jackson annotations for enum

mcassidygamma:
I encountered an issue when using the jaxrs-spec generator. For a pojo which used an enum, the enum was trying to use @JsonCreator and @JsonValue which were never being imported, thus compilation would fail.

I've added the imports and added a test also.
Resubmitting after grabbing all the most recent master branch changes.


blaukvik:
This would be nice - 
Is there a SNAPSHOT of maven-plugin with this somewhere ?  I can test  :-)

Issue #678, opened 2020-05-13
Wrong path generated in AbstractJavaJAXRSServerCodegen.java


Maxouwell:
The commit https://github.com/swagger-api/swagger-codegen-generators/commit/82b624b911ef4137b485af02553e165240a7e983 modified the DefaultCodegenConfig.java class in the apiFilename function
```
public String apiFilename(String templateName, String tag) {
        String suffix = apiTemplateFiles().get(templateName);
OLD        return apiFileFolder() + '/' + toApiFilename(tag) + suffix;
NEW        return apiFileFolder() + File.separator + toApiFilename(tag) + suffix;
    }
```
It potentielly break all the inheriting childrens which search '/' in the path.
One example is AbstractJavaJAXRSServerCodegen.java :
```
public String apiFilename(String templateName, String tag) {
        String result = super.apiFilename(templateName, tag);

        if ( templateName.endsWith("Impl.mustache") ) {
            int ix = result.lastIndexOf('/');
            result = result.substring(0, ix) + "/impl" + result.substring(ix, result.length() - 5) + "ServiceImpl.java";
            result = result.replace(apiFileFolder(), implFileFolder(implFolder));
        } else if ( templateName.endsWith("Factory.mustache") ) {
            int ix = result.lastIndexOf('/');
            result = result.substring(0, ix) + "/factories" + result.substring(ix, result.length() - 5) + "ServiceFactory.java";
            result = result.replace(apiFileFolder(), implFileFolder(implFolder));
        } else if ( templateName.endsWith("Service.mustache") ) {
            int ix = result.lastIndexOf('.');
            result = result.substring(0, ix) + "Service.java";
        }
        return result;
    }
```
The lastIndexOf search should be done on File.separator.
Alternatively, the DefaultCodegenConfig commit can be reverted.

HugoMario:
fixed by https://github.com/swagger-api/swagger-codegen-generators/pull/1053

PR #1053, opened 2022-08-02
Update AbstractJavaJAXRSServerCodegen.java

PippoBaudoCorporation:
the generation of the code will fail on windows beacause the separator is \ using File.separatorChar is fine for both os


bmsq:
Its worth mentioning that this has been a problem for quite some time and would resolve the following issues:
#678, [swagger-codegen #9045](https://github.com/swagger-api/swagger-codegen/issues/9045)

HugoMario:
thanks a lot @PippoBaudoCorporation for this PR

Issue #1066, opened 2022-09-21
[Java] bugs in AbstractJavaCodegen.fixUpParentAndInterfaces() method


AttilaW-Keytiles:
There are multiple bugs in io.swagger.codegen.v3.generators.java.AbstractJavaCodegen.fixUpParentAndInterfaces(CodegenModel, Map<String, CodegenModel>) method.
* The `CodegenModel parentModel = codegenModel.parentModel;` line should be moved into the for cycle `for (CodegenProperty codegenProperty : codegenModel.vars)` - because this way if the 1st property in the array does not have a conflict then it will "walk up" to `parentModel = null` and when the 2nd property comes in order to be checked - even if it has a conflict - would be never detected but skipped due to the null check condition
* The method if evaluates to `hasConflict = true` then does not change the `codegenProperty.nameInCamelCase` field at all  - which is used in mustache templates (generating addXXXItem() methods for example)

AttilaW-Keytiles:
I created a test case for this in AbstractJavaCodegenTest.java (to prove the bug) and also a fix for it in the code
Wanted to open a PR. But can not push up the bug branch I created - it's denied. Something I must do wrong however I tried to do everything as I read in the contributing note... :-(
If someone has a clue what do I do wrong I'm happy to shoot for the PR!

AttilaW-Keytiles:
OK I figured it out how PRs work - PR is in place :-)

PR #1067, opened 2022-09-25
fixing Issue #1066 - added unit test to prove the failure and the code fix

AttilaW-Keytiles:


AttilaW-Keytiles:
something is wrong so I close it (sorry, first time trying.. :-) )

PR #1068, opened 2022-09-25
fixing Issue #1066

AttilaW-Keytiles:
added unit test to prove the failure and the code fix


AttilaW-Keytiles:
still something is not good lets close

PR #1069, opened 2022-09-25
fix for AbstractJavaCodegen.fixUpParentAndInterfaces method

AttilaW-Keytiles:
fixing Issue #1066 - added unit test to prove the failure and the code fix


frantuma:
Thanks!

PR #1080, opened 2022-11-13
Update formParams.mustache

Alpcom:
This should be a simple miss but when you have multiple file parts the name is in conflict


Alpcom:
related issue : https://github.com/swagger-api/swagger-codegen-generators/issues/1081

PR #1100, opened 2023-01-30
avoid double pattern property assignment in order to fix issue

HugoMario:


HugoMario:
fixes https://github.com/swagger-api/swagger-codegen/issues/9509

PR #1143, opened 2023-05-03
fixed model name prefix/suffix issues

kapcsandi:
Fixed model name prefix/suffix generator issues where the prefix/suffix can be duplicated in the api/controller client codes.
- fixes swagger-api/swagger-codegen#7865
- fixes swagger-api/swagger-codegen#7866


kapcsandi:
@HugoMario can you please review this commit?

PR #1184, opened 2023-09-05
set discriminator property name for all schemas

### LINKED ISSUE #1183: [java][okhttp4-gson] ApiClient.mustache Generates invalid code in Baseclass
Basically same issue as in https://github.com/swagger-api/swagger-codegen-generators/issues/617 imho.
When using an enum as a discriminator in an inheritance hierarchy, the okhttp4-gson code generates an invalid java object.
As far as i can tell the cause is the following part in pojo.mustache:
`{#gson}}
  {{#discriminator}}
  public {{classname}}() {
    this.{{discriminator.propertyName}} = this.getClass().getSimpleName(); // HERE this would assign to an enum a string
  }
  {{/discriminator}}
  {{/gson}}
`
This very issue existed in the 2.x codebase and has been fixed here: https://github.com/swagger-api/swagger-codegen/pull/9207/files
And especially this: https://github.com/swagger-api/swagger-codegen/pull/9983/files
As far as I can tell, but unfortunately, this has been reintroduced in the 3.x codebase for okhttp4-gson.
I know this worked in the past, since we're about to migrate our old generated cliient to okhttp4-gson, which fails to be generated with the new 3.x codebase.
It still works for jersey3, which is probably our workaround.

HugoMario:
fix #1183

### DIFF
--- a/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
+++ b/src/main/java/io/swagger/codegen/v3/generators/DefaultCodegenConfig.java
@@ -1344,6 +1344,9 @@ public abstract class DefaultCodegenConfig implements CodegenConfig {
         codegenModel.getVendorExtensions().put(CodegenConstants.IS_ALIAS_EXT_NAME, typeAliases.containsKey(name));
 
         codegenModel.discriminator = schema.getDiscriminator();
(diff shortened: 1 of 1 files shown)


chkpnt:
This might have introduced a [regression](https://github.com/swagger-api/swagger-codegen/issues/12335).

PR #1245, opened 2024-01-11
Delete src/main/resources/logback.xml

### LINKED ISSUE #1015: logback.xml not excluded in final jar artifact
Hello,

is there a particular reason to not exclude logback.xml from final jar artifact like it is done with swagger-codegen for instance

https://github.com/swagger-api/swagger-codegen/blob/3.0.0/modules/swagger-codegen/pom.xml
```
        <resources>
            <resource>
                <directory>src/main/resources</directory>
                <excludes>
                    <exclude>logback.xml</exclude>
                </excludes>
            </resource>
        </resources>
```

I use it as a dependency ( maybe I am wrong ) and I suspect it changed our default logging configuration

https://github.com/swagger-api/swagger-codegen-generators/blob/master/src/main/resources/logback.xml
```
    <root level="error">
        <appender-ref ref="STDOUT"/>
    </root>
```

sirlatrom:
Fixes #1015.


HugoMario:
Hello @sirlatrom 
#1015 is proposing to exclude it from artifact when the project is build/released. but here you are removing the file.

sirlatrom:
I'm fine with either solution, as long as a release is made which fixes the issue. Feel free to close this PR accordingly.

PR #1291, opened 2024-05-27
Java - fix @NotNull resolving from required/nullable

frantuma:


yeikel:
@frantuma It seems that this change introduced this https://github.com/swagger-api/swagger-codegen-generators/issues/1295

PR #1305, opened 2024-08-19
Not undefined annotation add

micryc:
This PR creates custom validator to check if any field with nullable and required set to true is not undefined


frantuma:
included and replaced by #1308

PR #1320, opened 2024-11-04
Use ParameterMessageInterpolation 

micryc:
refs-https://github.com/swagger-api/swagger-codegen/issues/12448
Use ParameterMessageInterpolation to get rid of error with hibernate missing EL dependency


frantuma (changes requested):
This looks ok but please rename commit adding ref to related ticket swagger-api/swagger-codegen/issues/12448

Just as a note for future reference, the other option instead of adding `	`ParameterMessageInterpolator` is adding EL implementation to dependencies. This would mean downgrade `el-api` to 5x for `jakarta`, and is currently not needed as no EL expression is used in validation annotations.

frantuma (approved):

PR #1335, opened 2025-02-24
[Java] [Feign] Add support for Feign 13

lucasls:
As pointed out in the issue https://github.com/swagger-api/swagger-codegen/issues/12524, compilation fails for generated classes when Feign 13 and above are used.

The reason is version 13 adds an overload to the RetryableException constructor with the type Long instead of Date, and the value is passed as null causing the compiler to fail to resolve the constructor.

This PR applies https://github.com/swagger-api/swagger-codegen/pull/12525 for version 3.0

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 
     <groupId>io.swagger.codegen.v3</groupId>
     <artifactId>swagger-codegen-generators</artifactId>
-    <version>1.0.56</version>
+    <version>1.0.57-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <build>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -12,7 +12,7 @@
 
     <groupId>io.swagger.codegen.v3</groupId>
     <artifactId>swagger-codegen-generators</artifactId>
-    <version>1.0.56</version>
+    <version>1.0.57-SNAPSHOT</version>
```
lucasls, line 15: Not sure if I should have bumped it here, let me know if I should revert

lucasls:
Hey @frantuma, this part of the fix to support Feign 13, fixing the 3.0 generator

PR #1346, opened 2025-05-28
issue-9509 adding escaping support to patterns added with ref properties

evgeniimv:
as was already stated in https://github.com/swagger-api/swagger-codegen/issues/9509 

> Regex pattern like this '^\d{1,13}.\d{1,5}$' are not escape property and the generated java code does not compile.

tho, it was fixed in v2 of swagger-codegen
https://github.com/swagger-api/swagger-codegen-generators/pull/1100
https://github.com/swagger-api/swagger-codegen/pull/12038

it remains broken for swagger v3,

> any chance to get this work resumed and merged, specially into v3? Thanks.

In this change I'm addressing `pattern` escaping for properties constructed from references.


evgeniimv:
@frantuma Hey Francesco, may I ask you to check this one, once you have time.
 Much appreciate it :)

